PR #23, opened 2025-01-11
bump to maili

bxue-l2:
Upgrade to maili

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,63 +12,64 @@ hokulea-proof = { path = "crates/proof", version = "0.1.0", default-features = f
 # We use git dependencies instead of version dependencies because Kona is moving very fast right now
 # but publish infrequently (last was 2 weeks ago). We want to make sure to use the latest code
 # while we're still figuring out how to integrate with it.
-kona-mpt = { git = "https://github.com/anton-rs/kona", rev = "8f21eff5d48f2a7e72380c15e24a927799dd8175", default-features = false }
-kona-derive = { git = "https://github.com/anton-rs/kona", rev = "8f21eff5d48f2a7e72380c15e24a927799dd8175", default-features = false }
-kona-driver = { git = "https://github.com/anton-rs/kona", rev = "8f21eff5d48f2a7e72380c15e24a927799dd8175", default-features = false }
-kona-executor = { git = "https://github.com/anton-rs/kona", rev = "8f21eff5d48f2a7e72380c15e24a927799dd8175", default-features = false }
-kona-proof = { git = "https://github.com/anton-rs/kona", rev = "8f21eff5d48f2a7e72380c15e24a927799dd8175", default-features = false }
-kona-std-fpvm = { git = "https://github.com/anton-rs/kona", rev = "8f21eff5d48f2a7e72380c15e24a927799dd8175", default-features = false }
-kona-preimage = { git = "https://github.com/anton-rs/kona", rev = "8f21eff5d48f2a7e72380c15e24a927799dd8175", default-features = false }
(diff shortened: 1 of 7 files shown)


bin/client/justfile
```diff
@@ -69,7 +69,7 @@ run-client-native-against-devnet verbosity='' block_number='' rollup_config_path
   L1_BEACON_RPC="http://127.0.0.1:5052"
   L2_RPC="http://127.0.0.1:9545"
   ROLLUP_NODE_RPC="http://127.0.0.1:7545"
-  ROLLUP_CONFIG_PATH="{{justfile_directory()}}/../../optimism/.devnet/rollup.json"
+  ROLLUP_CONFIG_PATH="/home/ubuntu/op-main-repo/.devnet/rollup.json"
```
samlaf, line 72: why are you reverting to a path that only works on your machine? did the relative one not work?

Cargo.toml
```diff
@@ -12,63 +12,64 @@ hokulea-proof = { path = "crates/proof", version = "0.1.0", default-features = f
...
+alloy-rpc-client = { version = "0.9.2", default-features = false }
 alloy-primitives = { version = "0.8.14", default-features = false }
-alloy-node-bindings = { version = "0.8.0", default-features = false }
-alloy-transport-http = { version = "0.8.0", default-features = false }
-alloy-rpc-types-engine = { version = "0.8.0", default-features = false }
-alloy-rpc-types-beacon = { version = "0.8.0", default-features = false }
+alloy-node-bindings = { version = "0.9.2", default-features = false }
+alloy-transport-http = { version = "0.9.2", default-features = false }
+alloy-rpc-types-engine = { version = "0.9.2", default-features = false }
+alloy-rpc-types-beacon = { version = "0.9.2", default-features = false }
```
samlaf, line 48: Any reason you're using 0.9.2 instead of 0.9? Both of those will anyways end up with the latest 0.9.x version being used, but 0.9.2 *could* cause issues if we import another library that requires <0.9.2

samlaf (commented):
overall LGTM. why is this needed now? are we sure maili is going to be more stable than op-allow? are op-succint and kailua also going to be using maili? if not we might run into dependency hell issues

bin/client/justfile
bxue-l2: nah, just my local thing, will revert this line

Cargo.toml
bxue-l2: just copy op-rs/kona latest release https://github.com/op-rs/kona/blob/14c7de51d75d15a82cdc8138c04bec5da85617ad/Cargo.toml#L80

bxue-l2:
upgrade to maili would work. 

But I am fine to just find a version that works with op-succint or kailua

Cargo.toml
```diff
@@ -78,13 +78,13 @@ tracing = { version = "0.1.41", default-features = false }
...
-proptest = "1.5.0"
+proptest = "1.6.0"
 criterion = "0.5.1"
 
 # Serialization
 rkyv = "0.8.9"
-serde = { version = "1.0.215", default-features = false }
-serde_json = { version = "1.0.133", default-features = false }
+serde = { version = "1.0.217", default-features = false }
+serde_json = { version = "1.0.135", default-features = false }
```
samlaf, line 87:
```suggestion
serde = { version = "1.0", default-features = false }
serde_json = { version = "1.0", default-features = false }
```

samlaf: if we don't absolutely need those minimum versions, then don't specify them. makes it easier to get into dependency hell.

samlaf (approved):
Overall LGTM. Haven't looked at rust's package manager enough, but overall feel like we should very rarely specify minor versions as minimum requirements.

PR #28, opened 2025-01-24
Seperate crate for blob crypto

bxue-l2:
This PR creates a cryptography crate that contains std library. The crate 
- uses the latest rust-kzg-bn254 to verify the blob
- make blob witness into its own module, such that it can be used by both client and host

TODO in the next PR
- make resources/g1.point configurable via config
- have a run native-client option for the purpose of generating the witness


crates/cryptography/src/witness.rs
```diff
@@ -0,0 +1,135 @@
...
+            Bytes::copy_from_slice(blob),
+            Bytes::copy_from_slice(&commitment_bytes),
+            proof_bytes.into(),
+        );
+
+        Ok(())
+    }
+
+    /// helpful function to populate struct all together
+    pub fn write(&mut self, blob: Bytes, commitment: Bytes, proof: Bytes) {
```
samlaf, line 87: why is this helpful? When would you use this instead of push_witness? Is this only for testing? Can you create a constructor that takes 3 populates vecs instead or do you really need to call this function repeatedly?

crates/cryptography/src/witness.rs
```diff
@@ -0,0 +1,135 @@
...
+        self.commitments.push(commitment);
+        self.proofs.push(proof);
+        info!("added a blob");
+    }
+
+    /// verify all the blobs contained within the struct. It ensures the integrity of the
+    /// blob with respect the commitment. Using the batch verification method is the reduction
+    /// the size of MSM. For n blob of 4096 field element, instead of n MSM of 4096 points,
+    /// it would only require O(1) MSM of n points.
+    pub fn batch_verify(&self) -> bool {
```
samlaf, line 98:
```suggestion
    /// Verify all the blobs contained within the struct. It ensures the integrity of the
    /// blobs with respect to their commitments. Using batch verification instead of verifying each element one by one reduces
    /// the size of the MSM. For n blob of 4096 field element, instead of n MSM of 4096 points,
    /// only O(1) MSM of n points is required.
    pub fn batch_verify(&self) -> bool {
```

crates/cryptography/src/witness.rs
```diff
@@ -0,0 +1,135 @@
...
+        info!("lib_blobs len {:?}", self.eigenda_blobs.len());
```
samlaf, line 99: lib_blobs?

crates/cryptography/src/witness.rs
```diff
@@ -0,0 +1,135 @@
...
+
+        // transform to rust-kzg-bn254 inputs types
+        // TODO should make library do the parsing the return result
+        let lib_blobs: Vec<Blob> = self.eigenda_blobs.iter().map(|b| Blob::new(b)).collect();
+        let lib_commitments: Vec<G1Affine> = self
```
samlaf, line 104: seems like youre serializing to bytes in push_wiutness and then deserializing here. Can you just store the typed structs instead of bytes instead?

crates/cryptography/src/witness.rs
```diff
@@ -0,0 +1,135 @@
...
+        append_left_padded_biguint_be(&mut commitment_bytes, &commitment_x_bigint);
+        append_left_padded_biguint_be(&mut commitment_bytes, &commitment_y_bigint);
+
+        let mut proof_bytes = vec![0u8; 0];
+
+        let proof = kzg.compute_blob_proof(&input, &commitment)?;
+        let proof_x_bigint: BigUint = proof.x.into();
+        let proof_y_bigint: BigUint = proof.y.into();
+
+        append_left_padded_biguint_be(&mut proof_bytes, &proof_x_bigint);
```
samlaf, line 73: this is pretty ugly. At least pull out into a proper function that we can just upstream to rust-kzg-bn254 lib.

crates/cryptography/src/witness.rs
```diff
@@ -0,0 +1,135 @@
...
+            })
+            .collect();
+        batch::verify_blob_kzg_proof(&lib_blobs, &lib_commitments, &lib_proofs)
+            .map_err(|_| false)
+            .unwrap()
+    }
+}
+
+/// This function convert a BigUint into 32Bytes vector in big endian format
+pub fn append_left_padded_biguint_be(vec: &mut Vec<u8>, biguint: &BigUint) {
```
samlaf, line 130: hmmm doesn't feel like this function is the right abstraction. Why not just create proper serde-like functions that we can then upstream to rust-kzg-bn254?

Also this function prop shouldnt be public

bin/client/src/cached_eigenda_provider.rs
```diff
@@ -0,0 +1,83 @@
...
+use hokulea_eigenda::EigenDABlobProvider;
+use hokulea_proof::eigenda_provider::OracleEigenDAProvider;
+
+use kona_proof::errors::OracleProviderError;
+
+use alloc::sync::Arc;
+use rust_kzg_bn254::srs::SRS;
+use std::sync::Mutex;
+
+/// CachedOracleEigenDAProvider is a wrapper outside OracleEigenDAProvider. Its intended use
```
samlaf, line 18:
```suggestion
/// CachedOracleEigenDAProvider is a wrapper around OracleEigenDAProvider. Its intended use
```

bin/client/src/cached_eigenda_provider.rs
```diff
@@ -0,0 +1,83 @@
...
+/// case is to fetch all eigenda blobs received during the derivation pipeline. So that it
```
samlaf, line 19:
```suggestion
/// case is to fetch all eigenda blobs associated with the certs received during the derivation pipeline. So that it
```

bin/client/src/cached_eigenda_provider.rs
```diff
@@ -0,0 +1,83 @@
...
+/// is able to compute and cache the kzg witnesses, which can be verified inside ZKVM by checking
```
samlaf, line 20: Could we create a zkvm crate or somethign of that effect? Would be good to separate the structs and logic that is only used for zkvm vs those that are used for fraud proofs.

bin/client/src/cached_eigenda_provider.rs
```diff
@@ -0,0 +1,83 @@
...
+/// the point opening at the random Fiat Shamir evaluation index.
+#[derive(Debug, Clone)]
+pub struct CachedOracleEigenDAProvider<T: CommsClient> {
+    /// The preimage oracle client.
+    oracle: OracleEigenDAProvider<T>,
+    /// kzg proof witness
+    pub witness: Arc<Mutex<EigenDABlobWitness>>,
+    /// srs file
```
samlaf, line 28:
```suggestion
    /// srs points
```

crates/cryptography/README.md
```diff
@@ -0,0 +1,3 @@
+# `cryptography`
+
+This crate contains bn254 logics for generating kzg proof for either client or host. This crate uses STD.
```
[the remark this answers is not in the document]
bxue-l2: crates dir contains a list of crates, some are std some are not

it does both prove and verify,. But sg.

crates/proof/src/eigenda_provider.rs
```diff
@@ -100,27 +100,4 @@ impl<T: CommsClient + Sync + Send> EigenDABlobProvider for OracleEigenDAProvider
 
         Ok(blob.into())
     }
-
-    async fn get_element(&mut self, cert: &Bytes, element: &Bytes) -> Result<Bytes, Self::Error> {
```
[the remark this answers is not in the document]
bxue-l2: we don't need this

crates/cryptography/README.md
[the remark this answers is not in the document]
bxue-l2: call it kzg-crypto

crates/cryptography/src/witness.rs
```diff
@@ -0,0 +1,135 @@
...
+use rust_kzg_bn254::errors::KzgError;
+use rust_kzg_bn254::kzg::KZG;
+use rust_kzg_bn254::srs::SRS;
+use tracing::info;
+
+/// A data structure for storing blobs, the commitment of the blob and the proof
+/// that provides efficient verification between the commitment and the blob
+/// without actually commiting the blob
+#[derive(Debug, Clone, Default)]
+pub struct EigenDABlobWitness {
```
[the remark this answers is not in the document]
bxue-l2: kzg-crypti

crates/cryptography/src/lib.rs
```diff
@@ -0,0 +1,12 @@
...
+    missing_debug_implementations,
+    missing_docs,
+    unreachable_pub,
+    rustdoc::all
+)]
+#![deny(unused_must_use, rust_2018_idioms)]
+#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
+#![cfg_attr(not(test), warn(unused_crate_dependencies))]
+
+pub mod witness;
```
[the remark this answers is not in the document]
bxue-l2: sg

crates/cryptography/README.md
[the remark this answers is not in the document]
bxue-l2: renamed to compute_kzg_proof in separete PR

crates/cryptography/src/witness.rs
```diff
@@ -0,0 +1,135 @@
...
+            proofs: Vec::new(),
+        }
+    }
+
+    /// This function computes a witness for a eigenDA blob
+    /// nitro code <https://github.com/Layr-Labs/nitro/blob/14f09745b74321f91d1f702c3e7bb5eb7d0e49ce/arbitrator/prover/src/kzgbn254.rs#L141>
+    /// could refactor in the future, such that both host and client can compute the proof
+    pub fn push_witness(&mut self, blob: &[u8], srs: &SRS) -> Result<(), KzgError> {
+        // TODO check if KZG can just take a reference to srs, it saves memory
+        let mut kzg = KZG::new(srs.clone());
```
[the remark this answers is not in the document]
bxue-l2: using a new pattern

` let srs = SRS::new("resources/g1.32mb.point", 268435456, 1024).unwrap();
    let mut kzg = KZG::new();

    let input = Blob::new(blob);
    let input_poly = input.to_polynomial_eval_form();

    kzg.calculate_and_store_roots_of_unity(blob.len() as u64).unwrap();

    let mut commitment_bytes = vec![0u8; 0];

    let commitment = kzg.commit_eval_form(&input_poly, &srs)?;`

crates/cryptography/src/witness.rs
bxue-l2: removed

crates/cryptography/src/witness.rs
bxue-l2: but not a good name

crates/cryptography/src/witness.rs
bxue-l2: changed

crates/cryptography/src/witness.rs
bxue-l2: the ideal is to ustream to rust-kzg-bn254

bin/client/src/cached_eigenda_provider.rs
bxue-l2: fixed

bin/client/src/lib.rs
```diff
@@ -51,6 +57,14 @@ where
     let beacon = OracleBlobProvider::new(oracle.clone());
     let eigenda_blob_provider = OracleEigenDAProvider::new(oracle.clone());
 
+    let eigenda_blob_witness = Arc::new(Mutex::new(EigenDABlobWitness::new()));
+    let srs = SRS::new("resources/g1.point", 268435456, 1024).unwrap();
```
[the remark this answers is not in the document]
bxue-l2: right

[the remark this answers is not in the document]
bxue-l2: 268435456 is number of SRS points, the only usecase is that you cannot load more points than there existed. 
268435456 should be hard coded in the kzg lib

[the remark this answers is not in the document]
bxue-l2: https://github.com/Layr-Labs/rust-kzg-bn254/issues/46

bin/client/src/cached_eigenda_provider.rs
```diff
@@ -0,0 +1,83 @@
...
+            return Err(OracleProviderError::Preimage(PreimageOracleError::Other(
+                e.to_string(),
+            )));
+        };
+
+        let last_commitment = witness.commitments.last().unwrap();
+
+        // make sure locally computed proof equals to returned proof from the provider
+        if last_commitment[..32] != cert_blob_info.blob_header.commitment.x[..]
+            || last_commitment[32..64] != cert_blob_info.blob_header.commitment.y[..]
```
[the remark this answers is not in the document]
bxue-l2: will consider later

PR #32, opened 2025-02-19
feat: add v2 cert in client

bxue-l2:
This PR does the following:
- allows derivation pipeline to multiplex among different cert version. 
- changes the get_blob, get_blob_v2 trait to return Blob.
- implement EigenDABlobProvider for PreloadedEigenDABlobProvider

Address work 1 from #31

Work to be done in separate PR
- Decide what is the key when accessing each field element

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,7 @@ rust-kzg-bn254-prover = { git = "https://github.com/Layr-Labs/rust-kzg-bn254", r
 rust-kzg-bn254-verifier = { git = "https://github.com/Layr-Labs/rust-kzg-bn254", rev = "b3e532e9aad533009849755d5ad7b9578a16bfb2", default-features = false }
 
 # EigenDA v2 struct
-eigenda-v2-struct-rust = { git = "https://github.com/bxue-l2/eigenda-v2-struct-rust", rev = "cdeaf4e6ed7c3d55e70bf97c2c04a5e056b780ef" }
+eigenda-v2-struct-rust = { git = "https://github.com/bxue-l2/eigenda-v2-struct-rust", rev = "d3e3b956a59dcd2bb6bb7ea01c9e9d699350c4d8" }
 
 ark-bn254 = "0.5.0"
 ark-ff = { version = "0.5.0", features = ["parallel"] }
--- a/crates/eigenda/src/eigenda.rs
+++ b/crates/eigenda/src/eigenda.rs
@@ -3,12 +3,13 @@
 
 use crate::eigenda_blobs::EigenDABlobSource;
 use crate::traits::EigenDABlobProvider;
-use crate::{BlobInfo, STALE_GAP};
+use crate::{BlobInfo, CertVersion, STALE_GAP};
 use alloy_rlp::Decodable;
 
 use alloc::{boxed::Box, fmt::Debug};
 use alloy_primitives::Bytes;
 use async_trait::async_trait;
+use eigenda_v2_struct_rust::EigenDAV2Cert;
 use kona_derive::{
     errors::{PipelineError, PipelineErrorKind},
     sources::EthereumDataSource,
@@ -61,30 +62,55 @@ where
     async fn next(&mut self, block_ref: &BlockInfo) -> PipelineResult<Self::Item> {
         // then acutally use ethereum da to fetch. items are Bytes
         let cert = self.ethereum_source.next(block_ref).await?;
-
-        // TODO add logics to multiplex depends on the cert version
-
-        // verify if cert is too stale
-        let cert_blob_info = BlobInfo::decode(&mut &cert.as_ref()[4..]).unwrap();
-        info!("cert_blob_info {:?}", cert_blob_info);
-        let rbn = cert_blob_info
-            .blob_verification_proof
-            .batch_medatada
-            .batch_header
-            .reference_block_number as u64;
         let l1_block_number = block_ref.number;
 
-        // check staleness
-        // TODO: this would require the op-rollup to follow the same pattern
-        // but passing blockId to proxy which implement the logic,
-        // see https://github.com/ethereum-optimism/optimism/blob/0bb2ff57c8133f1e3983820c0bf238001eca119b/op-alt-da/damgr.go#L211
-        if rbn + STALE_GAP < l1_block_number {
-            // TODO: double check
-            return Err(PipelineErrorKind::Temporary(PipelineError::EndOfSource));
-        }
+        let cert_version_byte = cert.as_ref()[3];
+        let cert_version: CertVersion = cert_version_byte.into();
+        match cert_version {
+            CertVersion::Version1 => {
+                // TODO if punctuality is checked elsewhere, then we don't need to deserialize here
+                let cert_blob_info = BlobInfo::decode(&mut &cert.as_ref()[4..]).unwrap();
+                info!("cert_blob_info {:?}", cert_blob_info);
+                let rbn = cert_blob_info
+                    .blob_verification_proof
+                    .batch_medatada
+                    .batch_header
+                    .reference_block_number as u64;
 
-        let eigenda_blob = self.eigenda_source.next(&cert).await?;
-        Ok(eigenda_blob)
+                // check staleness
+                // TODO: this would require the op-rollup to follow the same pattern
+                // but passing blockId to proxy which implement the logic,
+                // see https://github.com/ethereum-optimism/optimism/blob/0bb2ff57c8133f1e3983820c0bf238001eca119b/op-alt-da/damgr.go#L211
+                if rbn + STALE_GAP < l1_block_number {
+                    // TODO: double check
+                    return Err(PipelineErrorKind::Temporary(PipelineError::EndOfSource));
+                }
+
+                let eigenda_blob = self.eigenda_source.next(&cert).await?;
+                Ok(eigenda_blob)
+            },
+            CertVersion::Version2 => {
+                // TODO if punctuality is checked elsewhere, then we don't need to deserialize here
+                let eigenda_v2_cert = match EigenDAV2Cert::decode(&mut &cert.as_ref()[4..]){
+                    Ok(c) => c,
+                    Err(_e) => return Err(PipelineErrorKind::Temporary(PipelineError::EndOfSource)),
+                };
+                let rbn = eigenda_v2_cert.batch_header_v2.reference_block_number as u64;
+                // check staleness
+                // TODO: this would require the op-rollup to follow the same pattern
+                // but passing blockId to proxy which implement the logic,
+                // see https://github.com/ethereum-optimism/optimism/blob/0bb2ff57c8133f1e3983820c0bf238001eca119b/op-alt-da/damgr.go#L211
+                if rbn + STALE_GAP < l1_block_number {
+                    // TODO: double check
+                    return Err(PipelineErrorKind::Temporary(PipelineError::EndOfSource));
+                }
+                let eigenda_blob = self.eigenda_source.next(&cert).await?;
+                Ok(eigenda_blob)
+            },
+            CertVersion::Unknown => {
+                return Err(PipelineErrorKind::Temporary(PipelineError::EndOfSource));
+            },
+        }
     }
 
     fn clear(&mut self) {
--- a/crates/eigenda/src/eigenda.rs
+++ b/crates/eigenda/src/eigenda.rs
@@ -88,12 +88,14 @@ where
 
                 let eigenda_blob = self.eigenda_source.next(&cert).await?;
                 Ok(eigenda_blob)
-            },
+            }
             CertVersion::Version2 => {
                 // TODO if punctuality is checked elsewhere, then we don't need to deserialize here
-                let eigenda_v2_cert = match EigenDAV2Cert::decode(&mut &cert.as_ref()[4..]){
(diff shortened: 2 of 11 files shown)


crates/eigenda/src/eigenda.rs
```diff
@@ -61,28 +62,57 @@ where
...
-        if rbn + STALE_GAP < l1_block_number {
-            // TODO: double check
-            return Err(PipelineErrorKind::Temporary(PipelineError::EndOfSource));
-        }
+        let cert_version_byte = cert.as_ref()[3];
+        let cert_version: CertVersion = cert_version_byte.into();
+        match cert_version {
+            CertVersion::Version1 => {
+                // TODO if punctuality is checked elsewhere, then we don't need to deserialize here
+                let cert_blob_info = BlobInfo::decode(&mut &cert.as_ref()[4..]).unwrap();
```
bxue-l2, line 72: where do we decide on the cert check

crates/proof/src/eigenda_provider.rs
```diff
@@ -100,4 +104,75 @@ impl<T: CommsClient + Sync + Send> EigenDABlobProvider for OracleEigenDAProvider
...
+            .length as usize;
+
+        // data_length measurs in field element, multiply to get num bytes
+        let mut blob: Vec<u8> = vec![0; blob_length * BYTES_PER_FIELD_ELEMENT];
+
+        // TODO: Investigate of using cert_rlp_bytes as key, instead of 96 bytes
+        let mut blob_key = [0u8; 96];
+
+        // the common key
+        let x: [u8; 32] = cert
```
bxue-l2, line 133: later create issue to figure key

Cargo.toml
```diff
@@ -76,7 +76,7 @@ rust-kzg-bn254-prover = { git = "https://github.com/Layr-Labs/rust-kzg-bn254", r
 rust-kzg-bn254-verifier = { git = "https://github.com/Layr-Labs/rust-kzg-bn254", rev = "b3e532e9aad533009849755d5ad7b9578a16bfb2", default-features = false }
 
 # EigenDA v2 struct
-eigenda-v2-struct-rust = { git = "https://github.com/bxue-l2/eigenda-v2-struct-rust", rev = "cdeaf4e6ed7c3d55e70bf97c2c04a5e056b780ef" }
+eigenda-v2-struct-rust = { git = "https://github.com/bxue-l2/eigenda-v2-struct-rust", rev = "d3e3b956a59dcd2bb6bb7ea01c9e9d699350c4d8" }
```
samlaf, line 79: can we work on moving all the rust repos to our monorepo next week?

samlaf: or create an eigenda-rs repo like op if we don't want all the rust stuff in the same monorepo (but I don't have a good reason why we wouldnt)

crates/eigenda/src/version.rs
```diff
@@ -0,0 +1,23 @@
+#[derive(Debug, PartialEq, Copy, Clone)]
+/// Represents the cert version derived from rollup inbox
+/// The version is needed to decode the Cert from serialiezd bytes
+/// Once a valid blob is retrieved, both versions use the identical
+/// logic to derive the rollup channel frame from eigenda blobs
+pub enum CertVersion {
+    /// unknown
+    Unknown,
```
samlaf, line 8: do we really need this? Think this is only typical in protobuf enums, for a reason that I still don't understand. Don't think its needed in rust though. Anything unknown should just return an error when we try to parse it to a correct version

crates/eigenda/src/version.rs
```diff
@@ -0,0 +1,23 @@
...
+    /// existing eigenda cert version
+    Version1,
+    /// lastest eigenda cert version
```
samlaf, line 11:
```suggestion
    /// latest eigenda cert version
```

crates/eigenda/src/version.rs
```diff
@@ -0,0 +1,23 @@
...
+    Version2,
+}
+
+impl From<u8> for CertVersion {
+    fn from(value: u8) -> Self {
+        match value {
+            0 => Self::Version1,
+            1 => Self::Version2,
+            _ => Self::Unknown,
```
samlaf, line 20: return an error here instead

crates/proof/src/eigenda_provider.rs
```diff
@@ -100,4 +104,75 @@ impl<T: CommsClient + Sync + Send> EigenDABlobProvider for OracleEigenDAProvider
 
         Ok(blob.into())
     }
+
+    /// get_blob_v2 takes a v2 cert type as opposed to bytes stream
+    async fn get_blob_v2(&mut self, cert: &EigenDAV2Cert) -> Result<Blob, Self::Error> {
+        let mut cert_rlp_bytes = Vec::<u8>::new();
+        // rlp encode of cert
+        cert.encode(&mut cert_rlp_bytes);
```
samlaf, line 112: I really really don't like these mutable things we're doing. It feels very c-like. Can we refactor .encode() to return a Vec instead? Unless we really really need the memory optimization because we're reusing vec buffers, which doesn't seem to be the case here though.

crates/proof/src/eigenda_provider.rs
```diff
@@ -100,4 +104,75 @@ impl<T: CommsClient + Sync + Send> EigenDABlobProvider for OracleEigenDAProvider
...
+
+        self.oracle
+            .write(&ExtendedHintType::EigenDACommitment.encode_with(&[&cert_rlp_bytes]))
+            .await
+            .map_err(OracleProviderError::Preimage)?;
+
+        let blob_length = cert
```
samlaf, line 119:
```suggestion
        let blob_length_symbols = cert
```

samlaf: Made Austin change all lengths to use suffix in core repo. Think we should adopt this convention everywhere. I personally prefer to use fe (field_element) instead of symbols, but symbols seems to be what most of our codebase is using.

crates/proof/src/eigenda_provider.rs
```diff
@@ -100,4 +104,75 @@ impl<T: CommsClient + Sync + Send> EigenDABlobProvider for OracleEigenDAProvider
...
+            .blob_inclusion_info
+            .blob_certificate
+            .blob_header
+            .commitment
+            .commitment
+            .y
+            .to_be_bytes();
+
+        blob_key[..32].copy_from_slice(&x);
+        blob_key[32..64].copy_from_slice(&y);
```
samlaf, line 151: can we encapsulate this in a function like `blob_key_prefix(cert)` or something?

Also is `blob_key` used by op elsewhere? It's kind of an unfortunate name clash with what eigenda core calls a blob_key...... would be good to use another name here if we can.

crates/proof/src/eigenda_provider.rs
```diff
@@ -100,4 +104,75 @@ impl<T: CommsClient + Sync + Send> EigenDABlobProvider for OracleEigenDAProvider
 
         Ok(blob.into())
     }
+
+    /// get_blob_v2 takes a v2 cert type as opposed to bytes stream
+    async fn get_blob_v2(&mut self, cert: &EigenDAV2Cert) -> Result<Blob, Self::Error> {
```
samlaf, line 109: add documentation for how cert is converted to blob_key and what blob_key's encoding is.

crates/proof/src/eigenda_provider.rs
```diff
@@ -100,4 +104,75 @@ impl<T: CommsClient + Sync + Send> EigenDABlobProvider for OracleEigenDAProvider
...
+                    &mut field_element,
+                )
+                .await
+                .map_err(OracleProviderError::Preimage)?;
+
+            // if field element is 0, it means the host has identified that the data
+            // has breached eigenda invariant, i.e cert is valid
+            if field_element.is_empty() {
+                return Err(OracleProviderError::Preimage(PreimageOracleError::Other(
+                    "field elememnt is empty, breached eigenda invariant".into(),
```
samlaf, line 169:
```suggestion
                    "field element is empty, breached eigenda invariant".into(),
```

crates/proof/src/eigenda_provider.rs
```diff
@@ -100,4 +104,75 @@ impl<T: CommsClient + Sync + Send> EigenDABlobProvider for OracleEigenDAProvider
...
+            self.oracle
+                .get_exact(
+                    PreimageKey::new(*keccak256(blob_key), PreimageKeyType::GlobalGeneric),
+                    &mut field_element,
+                )
+                .await
+                .map_err(OracleProviderError::Preimage)?;
+
+            // if field element is 0, it means the host has identified that the data
+            // has breached eigenda invariant, i.e cert is valid
```
samlaf, line 166:
```suggestion
            // has breached eigenda invariant, i.e cert is invalid
```

samlaf: ?

crates/proof/src/preloaded_eigenda_provider.rs
```diff
@@ -14,7 +18,7 @@ use tracing::info;
 #[derive(Clone, Debug, Default)]
 pub struct PreloadedEigenDABlobProvider {
```
samlaf, line 19: can you add documentation here? You've documented the invariants it respects, but not how its used. I'm guessing this is the guy that's preloaded from the first derivation pipeline run on the native client, and passed to the zkvm? Would be good to document this

crates/proof/src/preloaded_eigenda_provider.rs
```diff
@@ -62,6 +65,30 @@ impl From<EigenDABlobWitnessData> for PreloadedEigenDABlobProvider {
     }
 }
 
+#[async_trait]
+impl EigenDABlobProvider for PreloadedEigenDABlobProvider {
+    // TODO investigate if create a speical error type EigenDABlobProviderError
```
samlaf, line 70:
```suggestion
    // TODO investigate if create a special error type EigenDABlobProviderError
```

crates/proof/src/preloaded_eigenda_provider.rs
```diff
@@ -62,6 +65,30 @@ impl From<EigenDABlobWitnessData> for PreloadedEigenDABlobProvider {
...
+    type Error = OracleProviderError;
+
+    /// Fetches a blob for V1
+    async fn get_blob(&mut self, _cert: &Bytes) -> Result<Blob, Self::Error> {
+        unimplemented!()
+    }
+
+    /// Fetches a blob for V2 using preloaded data
+    /// Return an error if cert does not match the immeditate next item
```
samlaf, line 79:
```suggestion
    /// Return an error if cert does not match the immediate next item
```

samlaf: bruh we're going to have to add a spellchecker to CI lol

crates/proof/src/preloaded_eigenda_provider.rs
```diff
@@ -62,6 +65,30 @@ impl From<EigenDABlobWitnessData> for PreloadedEigenDABlobProvider {
...
 }
 
+#[async_trait]
+impl EigenDABlobProvider for PreloadedEigenDABlobProvider {
+    // TODO investigate if create a speical error type EigenDABlobProviderError
+    type Error = OracleProviderError;
+
+    /// Fetches a blob for V1
+    async fn get_blob(&mut self, _cert: &Bytes) -> Result<Blob, Self::Error> {
+        unimplemented!()
```
samlaf, line 75: why are we only implementing v2?

crates/eigenda/src/version.rs
bxue-l2: it is needed to handle version byte 3,4 which we don't define the cert structure and handling

Cargo.toml
bxue-l2: Fine, but Rust version is very unstable. Any good reason you want to move it now?

crates/proof/src/eigenda_provider.rs
bxue-l2: this is alloy rlp encode interface, https://docs.rs/alloy-rlp/latest/alloy_rlp/ . I didn't implement anything, but instead use derive

bxue-l2: I don't see the benefits to implement it on our own. But if you insists, you can return Vec

crates/proof/src/eigenda_provider.rs
bxue-l2: symbol is a term used for RS purpose, but fine for me

crates/proof/src/preloaded_eigenda_provider.rs
bxue-l2: because we don't intent to add secure integration of zk fault proof for v1 right now. But anyone interested can pick up

crates/eigenda/src/version.rs
bxue-l2: this is an enum, I think it is pretty un-customary, and not clean

crates/proof/src/eigenda_provider.rs
bxue-l2: TODO this is wrong thing to use as key, to do later

crates/proof/src/preloaded_eigenda_provider.rs
bxue-l2: Add comment to clarify "we dont plan to implemtn for v1"

crates/proof/src/eigenda_provider.rs
bxue-l2: addressed here https://github.com/Layr-Labs/hokulea/pull/40

PR #33, opened 2025-02-20
feat: v2 cert validity validation to hokulea client

bxue-l2:
The actual diff for this PR is quite small, huge bloat from lock file

This PR adds cert validity logics by introducing a new struct 
- CertValidity, containing claimed_validity and receipt that supports the claim
- in dev mode, we don't need receipt to support claim

This Pr adds CertValidityJournal, which is a struct emitted from zkvm view call proof
- see example here, https://github.com/bxue-l2/risc0-ethereum/blob/add-da-cert/examples/dacert-v2-verifier/methods/guest/src/bin/dacert_v2_verifier.rs#L134 and https://github.com/bxue-l2/risc0-ethereum/blob/add-da-cert/examples/dacert-v2-verifier/apps/src/bin/publisher.rs#L47

This Pr address the issue from 
- Implement validity validation logics from PR #31 

More Issue
- Need to have clarity on how to collect the correct view call function signature. At minimum, we need a router and try to keep func sig identical.


Cargo.toml
```diff
@@ -103,6 +103,9 @@ revm = { version = "16.0.0", default-features = false }
 # K/V database
 rocksdb = { version = "0.22.0", default-features = false }
 
+# ZKVM
+risc0-zkvm = { version = "1.2.3", features = ["heap-embedded-alloc", "unstable"] }
```
samlaf, line 107: think in general its more customary to turn off all features in the workspace cargo.toml and only turn on the ones that you need per crate. Although here I'm guessing this risc0-zkvm is only used by the proof crate right?

crates/proof/src/cert_validity.rs
```diff
@@ -0,0 +1,44 @@
...
+use eigenda_v2_struct_rust::EigenDAV2Cert;
+use risc0_zkvm::is_dev_mode;
+use risc0_zkvm::Receipt;
+
+#[derive(Debug, Clone, Default)]
+pub struct CertValidity {
+    /// the claim about if the cert is valid
+    pub claimed_validity: bool,
+    /// a zkvm proof attesting the above result
+    /// in dev mode, receipt is ignored
```
samlaf, line 11:
```suggestion
    /// A zkvm proof attesting the claimed_validity.
    /// In dev mode, receipt is ignored.
```

samlaf: Best to use full punctuation, otherwise it reads like "proof attesting the above result in dev mode", aka that the proof is only used in dev mode.

crates/proof/src/cert_validity.rs
```diff
@@ -0,0 +1,44 @@
...
+
+impl CertValidity {
+    /// verify if the receipt for cert is valid
+    /// note this is different from if the cert itself is valid as in the is_valid field
+    pub fn validate_cert_receipt(
+        &self,
+        eigenda_cert: &EigenDAV2Cert,
+        validity_call_verifier_id: B256,
+    ) {
+        if !is_dev_mode() {
```
samlaf, line 23: clearer to flip this around and just return early when in dev_mode

crates/proof/src/cert_validity.rs
```diff
@@ -0,0 +1,44 @@
...
+            use crate::journal::CertValidityJournal;
+            use alloy_rlp::Decodable;
+            use risc0_zkvm::sha::Digest;
```
samlaf, line 26: move this to top of file.

crates/proof/src/preloaded_eigenda_provider.rs
```diff
@@ -30,15 +30,16 @@ impl From<EigenDABlobWitnessData> for PreloadedEigenDABlobProvider {
...
-            ));
+                // TODO figure out a way to pass down validity_call_verifier_id
+                // at minimum, this value needs to come from system config from derivation
+                B256::default(),
+            );
 
             // if valid, check blob kzg integrity
-            let is_valid = value.validity_proofs[i].0;
-            if is_valid {
+            if value.validity[i].claimed_validity {
```
samlaf, line 42: think we can get rid of the CertValidity struct if `validate_cert_receipt` returns `validity` (which wouldn't be a claimed validity anymore, it would be the actual validity).

crates/proof/src/journal.rs
```diff
@@ -0,0 +1,22 @@
+use alloy_primitives::B256;
+use alloy_rlp::{RlpDecodable, RlpEncodable};
+use eigenda_v2_struct_rust::EigenDAV2Cert;
+
+/// CertValidityJournal is a data structure committed by zkvm guest code, that comes with
+/// corresponding zk proof attesting its validity
+#[derive(PartialEq, Eq, Ord, PartialOrd, Clone, Copy, Debug, RlpEncodable, RlpDecodable)]
+pub struct CertValidityJournal {
+    /// indicate if cert is honest
+    pub is_valid: bool,
```
samlaf, line 10: I'm confused. What's the difference between this struct and the other CertValidity struct? They both contain is_valid. Would be nice if we can find a way to simplify this redundancy/complexity

crates/proof/src/cert_validity.rs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+impl CertValidity {
+    /// verify if the receipt for cert is valid
+    /// note this is different from if the cert itself is valid as in the is_valid field
+    pub fn validate_cert_receipt(
+        &self,
+        eigenda_cert: &EigenDAV2Cert,
+        validity_call_verifier_id: B256,
+    ) {
```
samlaf, line 22: if you define this function on `Receipt` directly and make it return `is_valid`, then I think we can get rid of the CertValidity struct altogether.

samlaf (commented):
First pass. I'm a bit confused. I might be misunderstanding, but it feels like there's a way to simplify the code. I suggested removing CertValidity struct in some comments, let me know what you think

crates/proof/src/journal.rs
bxue-l2: CertValidityJournal is emitted by a smart contract, we need to check the claimed validity is consistent to the claim in the local derivation.

PR #34, opened 2025-02-20
feat: add v2 host

bxue-l2:
hint type and refactor fetcher

### DIFF
--- a/crates/eigenda/src/eigenda_data.rs
+++ b/crates/eigenda/src/eigenda_data.rs
@@ -1,4 +1,4 @@
-use crate::{BLOB_ENCODING_VERSION_0, BYTES_PER_FIELD_ELEMENT};
+use crate::BYTES_PER_FIELD_ELEMENT;
 use alloc::vec;
 use alloy_primitives::Bytes;
 use bytes::buf::Buf;
@@ -53,7 +53,7 @@ impl EigenDABlobData {
     ///
     /// The length of (header + payload) by the encode function is always multiple of 32
     /// The eigenda proxy does not take such constraint.
-    pub fn encode(rollup_data: &[u8]) -> Self {
+    pub fn encode(rollup_data: &[u8], blob_version: u8) -> Self {
         let rollup_data_size = rollup_data.len() as u32;
 
         // encode to become raw blob
@@ -69,7 +69,7 @@ impl EigenDABlobData {
 
         let mut raw_blob = vec![0u8; blob_size as usize];
 
-        raw_blob[1] = BLOB_ENCODING_VERSION_0;
+        raw_blob[1] = blob_version;
         raw_blob[2..6].copy_from_slice(&rollup_data_size.to_be_bytes());
 
         // encode length as uint32
@@ -85,6 +85,7 @@ impl EigenDABlobData {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::BLOB_ENCODING_VERSION_0;
     use alloc::vec;
     use alloy_primitives::Bytes;
     use kona_derive::errors::BlobDecodingError;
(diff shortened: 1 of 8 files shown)


crates/proof/src/hint.rs
```diff
@@ -78,6 +80,7 @@ impl From<ExtendedHintType> for &str {
     fn from(value: ExtendedHintType) -> Self {
         match value {
             ExtendedHintType::EigenDACommitment => "eigenda-commitment",
+            ExtendedHintType::EigenDACommitmentV2 => "eigenda-commitment-v2",
```
samlaf, line 83: Are these really commitments?
I've been trying to only use commitments for the bytes sent to the batcher-inbox: https://specs.optimism.io/experimental/alt-da.html#example-commitments
So iiuc the bytes would look like:
0x01 (altda) | 0x01 (generic_commitment) | 0x00 (eigenda) | 0x1 (eigendav1) | v1Cert
0x01 (altda) | 0x01 (generic_commitment) | 0x00 (eigenda) | 0x1 (eigendav2) | v2Cert

Is this also what you're thinking? Should this here be called EigenDACertV2 instead? or is it also passing the entire commitment bytes?

crates/proof/src/eigenda_provider.rs
```diff
@@ -112,7 +112,7 @@ impl<T: CommsClient + Sync + Send> EigenDABlobProvider for OracleEigenDAProvider
         cert.encode(&mut cert_rlp_bytes);
 
         self.oracle
-            .write(&ExtendedHintType::EigenDACommitment.encode_with(&[&cert_rlp_bytes]))
+            .write(&ExtendedHintType::EigenDACommitmentV2.encode_with(&[&cert_rlp_bytes]))
```
samlaf, line 115: encode_with's argument is called "data". Can we rename it to cert instead?

Also this convinced me "commitment" is the wrong word here. See my other comment.

samlaf (commented):
This PR is introducing support for v2 certs no??? This is awesome, the title should be something more serious like
`feat: add support for v2 certs`
instead of
`add hint type and refactor fetcher`

crates/eigenda/src/eigenda_data.rs
```diff
@@ -53,7 +53,7 @@ impl EigenDABlobData {
     ///
     /// The length of (header + payload) by the encode function is always multiple of 32
     /// The eigenda proxy does not take such constraint.
-    pub fn encode(rollup_data: &[u8]) -> Self {
+    pub fn encode(rollup_data: &[u8], blob_version: u8) -> Self {
```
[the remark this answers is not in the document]
bxue-l2: right, it is the cert version

crates/proof/src/hint.rs
bxue-l2: this is the cert. The commitment types are already extracted away. 

I like renaming to v1 and v2 cert. Why? we shouldn't have cert selector everywhere. It should only decide at one place

bxue-l2: https://github.com/Layr-Labs/hokulea/commit/e847a16faaae66713b6b639e733442684066db94

crates/proof/src/eigenda_provider.rs
bxue-l2: https://github.com/Layr-Labs/hokulea/commit/e847a16faaae66713b6b639e733442684066db94

bxue-l2:
> This PR is introducing support for v2 certs no??? This is awesome, the title should be something more serious like `feat: add support for v2 certs` instead of `add hint type and refactor fetcher`

This PR merges to #33 
PR #33 merges to #32. PR #33 deals with view call zkvm proof
PR #32 is the first PR that introduces the V2 cert

crates/proof/src/hint.rs
samlaf: agree with the reasoning, thanks for the change

crates/proof/src/eigenda_provider.rs
samlaf: can you also rename `encode_with`'s argument?

PR #41, opened 2025-03-04
feat: refactor hokulea host to recent kona host and upgrade dependency

bxue-l2:
This PR upgrades the dependency of hokulea and inevitably refactors the hokulea host to recent kona host.

The refactor has benefits of 
- staying closer to kona, and make it easy to integrate with others. 
- Remove boilerplate code previously needed because of missing abstraction

The following paragraphs describe the high level change

The old hokulea implementation requires
* OnlineHintRouter, containing
    * FetcherWithEigenDASupport, containing
        * fetcher
        * kv_store
        * eigenda_blob_provider
        * last_eigenda_hint
    * get_preimage 
        * Route hint by examining last_eigenda_hint to determine what fetch to use
        * If has eigenda_hint, then use get_preimage_eigenda
        * If not, use Kona fetcher
* OnlinePreimageFetcher
    * FetcherWithEigenDASupport
        * Call get_preimage implementation by FetcherWithEigenDASupport
* OfflinePreimageFetcher
    * KV
        * kv_store.get

New Kona implementation provides a nice abstraction
* HintHandler trait requires OnlineHostBackendCfg trait
* OnlineHostBackendCfg is determined by user to specify HintType and Providers
* In hokulea, we specify HintType as ExtendedHintType, Providers as SingleChainProvidersWithEigenDA

The new hokulea implementation requires
* Remove 
    * ExtendedHint, because HintType itself has abstraction over its type
    * OnlinePreimageFetcher, because provider is abstracted with trait via OnlineHostBackendCfg
    * OnlineHintRouter, because hint is abstracted with trait via OnlineHostBackendCfg
* Reuse OnlineHostBackend defined by Kona
* New struct
    * SingleChainHostWithEigenDA, that runs serveer
    * SingleChainProvidersWithEigenDA, impl Provider
    * SingleChainWithEigenDAHintHandle
        * A wrapper that routes fetch_hint among Kona and eigenda provider


bin/client/src/lib.rs
```diff
@@ -40,21 +48,26 @@ where
...
     // If the claimed L2 block number is less than the safe head of the L2 chain, the claim is
     // invalid.
-    let safe_head = fetch_safe_head(oracle.as_ref(), boot.as_ref(), &mut l2_provider).await?;
+    // Fetch the safe head's block header.
+    let safe_head = l2_provider
+        .header_by_hash(safe_head_hash)
+        .map(|header| Sealed::new_unchecked(header, safe_head_hash))?;
+
+    // If the claimed L2 block number is less than the safe head of the L2 chain, the claim is
+    // invalid.
```
samlaf, line 70: why? is it forcing us to make a claim about an unsafe block only? why?

bin/client/src/lib.rs
```diff
@@ -83,24 +96,41 @@ where
...
+        None,
+    );
+
     let mut driver = Driver::new(cursor, executor, pipeline);
 
     // Run the derivation pipeline until we are able to produce the output root of the claimed
     // L2 block.
-    let (number, output_root) = driver
-        .advance_to_target(&boot.rollup_config, Some(boot.claimed_l2_block_number))
+    let (safe_head, output_root) = driver
```
samlaf, line 132: why is this safe_head? isn't it advancing to the claimed_l2_block which is ahead of the safe_head?

bin/host/src/cfg.rs
```diff
@@ -0,0 +1,132 @@
...
+#[derive(Default, Parser, Serialize, Clone, Debug)]
+#[command(styles = cli_styles())]
+pub struct SingleChainHostWithEigenDA {
+    #[clap(flatten)]
+    pub kona_cfg: kona_host::single::SingleChainHost,
+
+    /// URL of the EigenDA RPC endpoint.
+    #[clap(long, env)]
+    #[arg(required = true)]
+    pub eigenda_proxy_address: Option<String>,
```
samlaf, line 28: why is it an Option if required=true?

bin/host/src/cfg.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub async fn start(self) -> Result<()> {
+        if self.kona_cfg.server {
+            let hint = FileChannel::new(FileDescriptor::HintRead, FileDescriptor::HintWrite);
+            let preimage =
+                FileChannel::new(FileDescriptor::PreimageRead, FileDescriptor::PreimageWrite);
+
+            self.start_server(hint, preimage).await?.await?
+        } else {
+            self.start_native().await
+        }
```
samlaf, line 42: is this code a copy-paste of some kona code? Might be good to add link just so we have the reference and can check if ever it gets updated that we match.

crates/proof/Cargo.toml
```diff
@@ -23,10 +23,11 @@ ark-ff.workspace = true
 alloy-primitives.workspace = true
 
 # Op Alloy
-op-alloy-rpc-types-engine.workspace = true
-op-alloy-genesis = { workspace = true, features = ["serde"] }
+# op-alloy-genesis = { workspace = true, features = ["serde"] }
```
samlaf, line 26: delete?

crates/proof/src/pipeline.rs
```diff
@@ -72,7 +73,7 @@ where
     /// Constructs a new oracle-backed derivation pipeline.
     pub fn new(
         cfg: Arc<RollupConfig>,
-        sync_start: PipelineCursor,
+        sync_start: Arc<RwLock<PipelineCursor>>,
         caching_oracle: Arc<O>,
```
samlaf, line 77: Is this also following a pattern from kona?
Would be good to add links everywhere we do this. This way its easier to know which pattern we're following from kona, and which constructions are new and specific to hokulea

Cargo.lock
samlaf, line 1: Can we add a `.gitattributes` file [like in eigenda](https://github.com/Layr-Labs/eigenda/blob/master/.gitattributes) to not have lockfiles count towards PR diff counts. This PR's diff isnt that big actually if you remove the lock changes.

bin/host/src/cfg.rs
```diff
@@ -0,0 +1,132 @@
...
+
+        Ok(task_handle)
+    }
+
+    /// Creates the providers with eigenda
+    pub async fn create_providers(&self) -> Result<SingleChainProvidersWithEigenDA> {
+        let kona_providers = self.kona_cfg.create_providers().await?;
+
+        let eigenda_blob_provider = OnlineEigenDABlobProvider::new_http(
+            //EIGENDA_ADDRESS.to_string(),
```
samlaf, line 88: remove?

bin/host/src/cfg.rs
```diff
@@ -0,0 +1,132 @@
...
+            self.eigenda_proxy_address.clone().unwrap(),
```
samlaf, line 89: you unwrap() to crash the program on an option. Why not just make the option not optional?

bin/host/src/handler.rs
```diff
@@ -0,0 +1,184 @@
...
+use kona_host::{single::SingleChainHintHandler, HintHandler, OnlineHostBackendCfg};
+
+use crate::cfg::SingleChainHostWithEigenDA;
+
+use kona_preimage::{PreimageKey, PreimageKeyType};
+use kona_proof::Hint;
+use tracing::trace;
+
+use eigenda_v2_struct_rust::EigenDAV2Cert;
+//use hokulea_compute_kzg_proof::compute_kzg_proof;
```
samlaf, line 17:
```suggestion
```

bin/host/src/handler.rs
```diff
@@ -0,0 +1,184 @@
+use alloy_primitives::keccak256;
+//use alloy_primitives::B256;
```
samlaf, line 2:
```suggestion
```

bin/host/src/handler.rs
```diff
@@ -0,0 +1,184 @@
...
+                )
+                .await?;
+            }
+        }
+        Ok(())
+    }
+}
+
+/// Fetch the preimage for the given hint and insert it into the key-value store.
+pub async fn fetch_eigenda_hint(
```
samlaf, line 61: should we make a separate function for v1 and v2?

Especially when we add the commitment logic, those will be separate for v1 and v2 right? So we'll have 2 switches in here? In general I prefer to #move-ifs-up as much as possible ;)

samlaf (commented):
Don't understand most of the changes happening here as its using new kona interfaces. Just highlighted some code nits. Otherwise LGTM

Cargo.toml
```diff
@@ -3,102 +3,106 @@ resolver = "2"
 members = ["bin/*", "crates/*"]
 
 [workspace.dependencies]
+kona-client = { git = "https://github.com/op-rs/kona", rev = "96aba6aa72053f628ac1047548530f27f7c9d25c", default-features = false, version = "0.1.0" }
+kona-host = { git = "https://github.com/op-rs/kona", rev = "96aba6aa72053f628ac1047548530f27f7c9d25c", default-features = false, features = ["single"], version = "0.1.0" }
+kona-derive = { git = "https://github.com/op-rs/kona", rev = "96aba6aa72053f628ac1047548530f27f7c9d25c", version = "0.2.3" }
```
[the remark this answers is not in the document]
bxue-l2: yes, kailua is on 96aba6aa72053f628ac1047548530f27f7c9d25c. Succinct is using its own fork of kona.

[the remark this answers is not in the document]
bxue-l2: I can try something later with a tagged kona vesion

bin/client/src/lib.rs
bxue-l2: this is kona's comment, the logic behind is the claimed l2 bn needs to be greater than the known safe one, cause there is no point of doing this

bin/host/src/cfg.rs
bxue-l2: good call

bin/host/src/cfg.rs
bxue-l2: good eye, copy is needed

bin/host/src/handler.rs
bxue-l2: i will to keep it there, in case we want it at some time

bin/host/src/handler.rs
bxue-l2: same as above

bin/client/src/lib.rs
bxue-l2: Name comes from kona. I think if advance_to_target call returns correctly, the state has transitioned to a safe head

bin/host/src/cfg.rs
bxue-l2: this is needed

crates/proof/src/pipeline.rs
bxue-l2: https://github.com/Layr-Labs/hokulea/pull/41/commits/b30b7d2d9adf32998964ce5cc4c1558bcef2ead3

Cargo.lock
bxue-l2: https://github.com/Layr-Labs/hokulea/pull/41/commits/b30b7d2d9adf32998964ce5cc4c1558bcef2ead3

bin/host/src/cfg.rs
bxue-l2: https://github.com/Layr-Labs/hokulea/pull/41/commits/952c7379d564b3d76033327c062baa33a7eb5a19

bin/host/src/cfg.rs
bxue-l2: https://github.com/Layr-Labs/hokulea/pull/41/commits/14760b158f72f332a74108f0d0fcef9a73f0563a

bin/host/src/cfg.rs
bxue-l2: https://github.com/Layr-Labs/hokulea/pull/41/commits/14760b158f72f332a74108f0d0fcef9a73f0563a

samlaf (approved):

PR #56, opened 2025-04-17
Rebase hokulea dependancies on kona, make it importable by op-succinct

bxue-l2:
This PR includes two parts
- rebase to kona
- remove parts that is incompatible to op-succinct because risc0 lib cannot be compiled into that
  - for example, CertValidity
- change codes to make it compatible to op-succinct, for example 
  - for example, Blob from rust-kzg-bn254 isn't serializable. Ideally we want it to be rkyvable.


Cargo.toml
samlaf, line 1: With the newest kona dependency, since they have bumped to msrv 1.85, then our tools no longer work. For example getting this in my vscode:
```
 The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.83.0 (5ffbef321 2024-10-29)).
  Consider trying a newer version of Cargo (this may require the nightly release).
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
```

If you rebase on top of master (I merged my PR which added rust version to mise.toml), we can then upgrade our rust version to 1.85. Feel like we should also add it to Cargo.toml, and specify MSRV in README, like kona does.

bin/host/src/main.rs
```diff
@@ -1,49 +1,19 @@
...
-        1 => LevelFilter::DEBUG,
-        _ => LevelFilter::TRACE,
-    };
-
-    let mut filter_builder = EnvFilter::builder()
-        .with_default_directive(level.into())
-        .parse("")?;
-
-    // Only show info logs for these http related crates.
-    // Their debug logs are extremely verbose, and clutter the output
```
samlaf, line 34: don't randomly remove my stuff lol. This is super useful to be able to  have dynamically configurable log level, and also filter a bunch of too verbose logs. Any reason why you reverted to using kona's init_tracing_subscriber? At the very least if you want to use their then we should add the EnvFilter I'm constructing here, instead of just passing None.

crates/compute-proof/src/steel/mod.rs
samlaf, line 1: Why are you removing this? Did you move it elsewhere? Or you just want to add it later once its actually working?

crates/eigenda/src/eigenda.rs
```diff
@@ -56,9 +56,15 @@ where
 {
     type Item = Bytes;
 
-    async fn next(&mut self, block_ref: &BlockInfo) -> PipelineResult<Self::Item> {
+    async fn next(
+        &mut self,
+        block_ref: &BlockInfo,
+        batcher_addr: Address,
+    ) -> PipelineResult<Self::Item> {
+        info!("EigenDADataSource next {} {}", block_ref, batcher_addr);
```
samlaf, line 64:
```suggestion
        info!("EigenDADataSource next block_ref={} batcher_addr={}", block_ref, batcher_addr);
```

crates/proof/src/cert_validity.rs
```diff
@@ -1,26 +1,28 @@
 use alloy_primitives::B256;
 use eigenda_v2_struct_rust::EigenDAV2Cert;
+use serde::{Deserialize, Serialize};
 
-use risc0_zkvm::Receipt;
+//use risc0_zkvm::Receipt;
```
samlaf, line 5:
```suggestion
```

crates/proof/src/cert_validity.rs
```diff
@@ -1,26 +1,28 @@
...
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
 pub struct CertValidity {
     /// the claim about if the cert is valid
     pub claimed_validity: bool,
     /// a zkvm proof attesting the above result
     /// in dev mode, receipt is ignored
     /// in the future, to make it generic for sp1-contract-call
-    pub receipt: Option<Receipt>,
+    pub receipt: Option<u8>, //Receipt
```
samlaf, line 14: what is this u8? Is this something op-succinct returns? What does it encode? What does it do?

crates/proof/src/cert_validity.rs
```diff
@@ -1,26 +1,28 @@
...
     /// verify if the receipt for cert is valid
     /// note this is different from if the cert itself is valid as in the is_valid field
     pub fn validate_cert_receipt(
         &self,
-        eigenda_cert: &EigenDAV2Cert,
-        validity_call_verifier_id: B256,
+        _eigenda_cert: &EigenDAV2Cert,
+        _validity_call_verifier_id: B256,
     ) {
+        /*
```
samlaf, line 25: why comment out this whole function?

crates/proof/src/eigenda_blob_witness.rs
```diff
@@ -3,17 +3,17 @@ use alloc::vec::Vec;
...
 
 /// One EigenDABlobWitnessData corresponds to one EigenDA cert
-#[derive(Debug, Clone, Default)]
+#[derive(Default, Debug, Clone, Serialize, Deserialize)]
 pub struct EigenDABlobWitnessData {
     /// eigenda v2 cert
     pub eigenda_certs: Vec<EigenDAV2Cert>,
     /// blob empty if cert is invalid
-    pub eigenda_blobs: Vec<Blob>,
+    pub eigenda_blobs: Vec<Vec<u8>>,
```
samlaf, line 16: Why are we using Vec<u8> for Blobs now? Is this forced because of op-succinct? Can we instead define our own types and then just typecast to op-succinct types right before the calls? Or can we ask them to properly type their stuff? Otherwise how do we expect this to work for both kailua and op-succinct?

crates/proof/src/pipeline.rs
```diff
@@ -90,18 +91,33 @@ where
...
+        pipeline
+            .signal(
+                ResetSignal {
+                    l2_safe_head,
+                    l1_origin: sync_start.read().origin(),
+                    system_config,
+                }
+                .signal(),
+            )
+            .await?;
```
samlaf, line 115: what is this doing? don't undersand the comment. What would be a proper way to handle this? How hard is it to implement?

Cargo.toml
```diff
@@ -3,24 +3,19 @@ resolver = "2"
...
-kona-executor = { git = "https://github.com/op-rs/kona", rev = "96aba6aa72053f628ac1047548530f27f7c9d25c", version = "0.2.3" }
-kona-mpt = { git = "https://github.com/op-rs/kona", rev = "96aba6aa72053f628ac1047548530f27f7c9d25c", features = [
-    "serde",
-], version = "0.1.2" }
-kona-preimage = { git = "https://github.com/op-rs/kona", rev = "96aba6aa72053f628ac1047548530f27f7c9d25c", features = [
-    "serde",
+kona-client = { git = "https://github.com/op-rs/kona", tag = "kona-client/v0.1.0-beta.16" }
+kona-host = { git = "https://github.com/op-rs/kona", tag = "kona-client/v0.1.0-beta.16" }
+kona-derive = { git = "https://github.com/op-rs/kona", tag = "kona-client/v0.1.0-beta.16", default-features = false }
+kona-driver = { git = "https://github.com/op-rs/kona", tag = "kona-client/v0.1.0-beta.16" }
```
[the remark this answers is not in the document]
bxue-l2: there is deps constraint on what op-succinct is using.

Cargo.toml
bxue-l2: sg

crates/compute-proof/src/steel/mod.rs
bxue-l2: op-succinct zkvm cannot take steel library. I think we should find out a way to separate them out. 
1. conditional compilation
2. put those zk proof generation inside the code region of op-succinct or kailua. Given they already the inherit the zkvm dependency, and clear separation

crates/proof/src/cert_validity.rs
bxue-l2: It is to do, I think we should make it opaque bytes string for compatibility. Same as above

crates/proof/src/eigenda_blob_witness.rs
bxue-l2: This is mostly about type serialization, Blob currently does not support serde, but it is needed in order to verify it inside zkvm. Yes, you need to pass the serialized bytes in.

I think we should make everything inside EigenDAV2Cert Rkyvable, it allows zero cost deserialization, but just another item on the todo list

crates/proof/src/pipeline.rs
bxue-l2: actually, this is a really good point

bin/host/src/main.rs
bxue-l2: ok, thought it was part of some legacy code, we copied from Kona, because it wasn't abstracted well enough before.

bxue-l2: 0a27bffc7b6cc5cfc1e36cb1f0208ca764071d9d

crates/proof/src/pipeline.rs
bxue-l2: 0a27bffc7b6cc5cfc1e36cb1f0208ca764071d9d

Cargo.toml
bxue-l2: rebased

Cargo.toml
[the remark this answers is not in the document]
samlaf: Think we should still add the version though.

crates/proof/src/pipeline.rs
samlaf: You left the comment
>         // TODO use proper handling like in kona to derive from l2_safe_head
Is that still needed? Or are we just copying what kona did (prob should mention that in a comment above the function itself instead?

crates/proof/src/eigenda_blob_witness.rs
bxue-l2: Created issue, https://github.com/Layr-Labs/rust-kzg-bn254/issues/53

PR #62, opened 2025-04-22
remove comment and merge conflict

bxue-l2:
This PR fixes some comment and fix CI because the container name has changed due to injection of chainId


samlaf (approved):
Fine with merging for the comment and memstore config change, but I prefer the justfile approach Ive taken in #63 which passes chainid as an argument to commands. I will rebase on top of this once you merge.

PR #65, opened 2025-04-23
chore: fix asterisc (no-std) build

### LINKED ISSUE #64: asterisc build fails
Once #63 merged, we can build the hokulea-client using `just build-client-for-asterisc`, the way same the kona-client can be built for asterisc.
However the build using kona's dockerfile environment currently fails for some weird dependency-hell types of errors. We need to investigate.

samlaf:
Attemps to fix #64 

Goal is to be able to build the client for the asterisc target, which is a no-std dependency, whereas client right now was depending on std features.

Ended up requiring a much larger change set than expected:
1. add default-features = false to every dependency is cargo.toml
2. refactored a lot of our crates to be `#![no-std]`
3. move witgen client from client-bin crate to its own create (its only used by op-succinct as a lib)
4. move https://github.com/bxue-l2/eigenda-v2-struct-rust into this repo as its own crate, so we can manage its dependency set (removed std feature from one of the deps, didnt seem used..)
5. updated rust-kzg-bn254 primitives+verifier crates to be no-std (see https://github.com/Layr-Labs/rust-kzg-bn254/pull/54, which we still haven't merged)

Now `just build-client-for-asterisc` works.


bxue-l2:
num-traits is no_std unless, it is marked without default feature

I think https://github.com/Layr-Labs/rust-kzg-bn254/blob/master/primitives/Cargo.toml#L17 didn't mark it?

Tried it, and you got a new error
Screenshot 2025-04-23 at 11 52 35 PM

bxue-l2:
Yes, I would like to merge it

bxue-l2:
OK, rust-kzg-bn254 is massive not no_std. See my change https://github.com/Layr-Labs/rust-kzg-bn254/commit/d7f821111be840b5fdfe4dcc7aea70a94d0d01af

this error
Screenshot 2025-04-24 at 12 19 33 AM

it imports ark_std

Cargo.toml
```diff
@@ -138,11 +125,8 @@ revm = { version = "19.5.0", default-features = false }
...
-opt-level = 3
-
-[profile.release]
-debug = 1
-lto = true
+# TODO: remove this patch before merging.
+[patch.crates-io]
+rust-kzg-bn254-primitives = { git = "https://github.com/Layr-Labs/rust-kzg-bn254.git", branch = "chore--make-asterisc-compile" }
+rust-kzg-bn254-prover = { git = "https://github.com/Layr-Labs/rust-kzg-bn254.git", branch = "chore--make-asterisc-compile" }
+rust-kzg-bn254-verifier = { git = "https://github.com/Layr-Labs/rust-kzg-bn254.git", branch = "chore--make-asterisc-compile" }
```
samlaf, line 132: TODO: first merge https://github.com/Layr-Labs/rust-kzg-bn254/pull/54, then release new version, and then update this PR before merging it

bxue-l2: Also have this in the release, https://github.com/Layr-Labs/rust-kzg-bn254/issues/53

bxue-l2: https://github.com/Layr-Labs/rust-kzg-bn254/pull/55

Can it merge to yours first? But I can rebase to master otherwise

crates/client/src/fp_client.rs
```diff
@@ -1,23 +1,23 @@
 extern crate alloc;
+use alloc::sync::Arc;
+use core::fmt::Debug;
+use core::prelude::v1::*;
```
bxue-l2, line 4: Do we need all of it

crates/eigenda-v2-struct/src/sol_struct.rs
```diff
@@ -0,0 +1,72 @@
+use alloy_sol_types::sol;
+
+sol! {
+    struct BatchHeaderV2 {
```
bxue-l2, line 4: We could just import it from solidity artifact

bxue-l2 (approved):
lgtm

crates/client/src/fp_client.rs
samlaf: probably not all, I can change to import them manually, but there was at least 5-6 of them.

crates/eigenda-v2-struct/src/sol_struct.rs
samlaf: we don't have the solidity artifact in hokulea do we..?

PR #68, opened 2025-04-30
chore(kt-devnet): add config file to run against holesky L1

samlaf:


bxue-l2:
@samlaf do you still want to merge it?

PR #73, opened 2025-04-30
use downgraded rust bn254 and expose witgen-crate

bxue-l2:


bxue-l2:
wait to see, if after sp1 upgrade, if libm is no longer a conflict

PR #79, opened 2025-05-05
bump to kona 1.0.0

bxue-l2:
This Pr bump kona deps to 1.0.0

Kona adds EvmFactory and op-revm to abstract execution and precompile, see https://github.com/op-rs/kona/issues/1543

### DIFF
--- a/bin/client/src/client.rs
+++ b/bin/client/src/client.rs
@@ -4,30 +4,30 @@ use alloc::sync::Arc;
 use core::fmt::Debug;
 
 use kona_client::single::FaultProofProgramError;
-use kona_executor::KonaHandleRegister;
 use kona_preimage::{HintWriterClient, PreimageOracleClient};
-use kona_proof::{l1::OracleBlobProvider, l2::OracleL2ChainProvider, CachingOracle};
+use kona_proof::{l1::OracleBlobProvider, CachingOracle};
 
 use hokulea_client::fp_client;
 use hokulea_proof::eigenda_provider::OracleEigenDAProvider;
+
+use alloy_evm::{EvmFactory, FromRecoveredTx, FromTxWithEncoded};
+use op_alloy_consensus::OpTxEnvelope;
+use op_revm::OpSpecId;
 
 /// The function uses the identical function signature as the kona client
 /// This is the basic hokulea client containing the minimal layer between kona client and hokulea host
 #[allow(clippy::type_complexity)]
 #[inline]
-pub async fn run_direct_client<P, H>(
+pub async fn run_direct_client<P, H, Evm>(
     oracle_client: P,
     hint_client: H,
-    _handle_register: Option<
-        KonaHandleRegister<
-            OracleL2ChainProvider<CachingOracle<P, H>>,
-            OracleL2ChainProvider<CachingOracle<P, H>>,
-        >,
-    >,
+    evm_factory: Evm,
 ) -> Result<(), FaultProofProgramError>
 where
     P: PreimageOracleClient + Send + Sync + Debug + Clone,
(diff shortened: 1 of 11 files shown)


bin/host/Cargo.toml
```diff
@@ -8,13 +8,15 @@ edition = "2021"
 hokulea-proof.workspace = true
 hokulea-client-bin.workspace = true
 hokulea-eigenda.workspace = true
+#hokulea-witgen-client.workspace = true
```
samlaf, line 11: delete

bin/host/src/cfg.rs
```diff
@@ -133,18 +134,25 @@ impl SingleChainHostWithEigenDA {
...
+        //let client_task = task::spawn(
+        //    hokulea_witgen_client::witgen_client::run_preloaded_eigenda_client(
+        //        OracleReader::new(preimage.client.clone()),
+        //        HintWriter::new(hint.client.clone()),
+        //        FpvmOpEvmFactory::new(
+        //            HintWriter::new(hint.client),
+        //            OracleReader::new(preimage.client),
+        //        ),
+        //    ),
+        //);
```
samlaf, line 147: when do you uncomment this? Could we put this behind an if statement, or a compile flag instead?

bin/client/src/client.rs
```diff
@@ -4,30 +4,30 @@ use alloc::sync::Arc;
...
-            OracleL2ChainProvider<CachingOracle<P, H>>,
-        >,
-    >,
+    evm_factory: Evm,
 ) -> Result<(), FaultProofProgramError>
 where
     P: PreimageOracleClient + Send + Sync + Debug + Clone,
     H: HintWriterClient + Send + Sync + Debug + Clone,
+    Evm: EvmFactory<Spec = OpSpecId> + Send + Sync + Debug + Clone + 'static,
+    <Evm as EvmFactory>::Tx: FromTxWithEncoded<OpTxEnvelope> + FromRecoveredTx<OpTxEnvelope>,
```
samlaf, line 30: is this associated type bound that we always add really needed? We don't even seem to be calling evmFactory ourselves. Guessing its used somewhere in some internal kona library, but I don't get it (not super familiar with how these bounds are used)

samlaf (approved):
Approved. kurtosis CI is failing because of https://github.com/ethpandaops/optimism-package/pull/243. Will fix in another PR

bin/host/src/cfg.rs
bxue-l2: I am thinking we should have a binary or test to just run preloaded eigenda client. 

test is best I think, will fix in another PR

bin/host/Cargo.toml
bxue-l2: I want to keep an appearance on where run_preloaded_eigenda_client can be used. 

It is a main entry point for intergration

bin/host/src/cfg.rs
bxue-l2: https://github.com/Layr-Labs/hokulea/issues/81

bin/host/Cargo.toml
bxue-l2: e629b9eccdb619c59895fb44c4f43df94e6eff8e

bin/client/src/client.rs
bxue-l2: I just make it consistent to kona library. Will think more about it, 
Maybe relate to this? https://github.com/op-rs/kona/pull/1400

samlaf (approved):
LGTM. Prob good to rebase on top of my kurtosis fix PR to make CI pass, but not strictly necessary.

PR #82, opened 2025-05-05
fix git attributes

bxue-l2:
let's see if it works


samlaf (approved):
I looked for this as well. I think they are equivalent, but not sure.

I think I it’s working as intended already actually. I think gitatteibutes just makes the files not show by default in PRs, but the line diffs still count.

PR #83, opened 2025-05-06
docs: add example demonstrating zk secure integration

bxue-l2:
This PR adds preloader example to demonstrate how to use some important data structure to securely integrate with zk.

The README contains the higher level logic for creating the witness and converting the witness into preloade eigenda blob provider.

This PR also rename the hokulea-witgen-client to hokulea-witgen, and move some functions out of the  hokulea-witgen-client. The motivation behind is that in the future we will have steel or sp1-contract-call be part of the witness generation. It will inevitably include those dependencies into the "hokulea-witgen". But now instead we could just put them into example. This is a better decoupling.

At last, the just file inside bin/client is moved to the top justfile, such that many logic can be reused to run the example/preloader

Btw, I changed the asterisc related justfile, they still didn't run.


bxue-l2:
Maybe a good idea to merge hokulea-compute-proof and hokulea-witgen

example/preloader/src/main.rs
```diff
@@ -37,27 +71,24 @@ where
...
+    // Run derivation for the first time to populate the witness data
+    info!("run_witgen_client: generating witness");
+    let wit = run_witgen_client(oracle.clone(), evm_factory.clone()).await?;
+
+    info!("done generating the witness");
+
+    // By this time,both Oracle and EigenDABlobWitnessData are generated by some party that runs run_witgen_client.
+    // This party now needs to send both of them as inputs to ZKVM. So imagine wit and oracle are sent away, and
+    // the code region below are some codes that runs inside ZKVM. The ZKVM will convert EigenDABlobWitnessData into
+    // a preloaded eigenda provider, that implements the trait get_blob. The run_fp_client are also run inside zkVM
```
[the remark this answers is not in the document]
samlaf:
```suggestion
    // At this point, the Oracle has been populated and the EigenDABlobWitnessData generated.
    // We now send both of them as inputs to the zkVM, which will convert the EigenDABlobWitnessData into
    // a PreloadedEigenDAProvider that implements the trait get_blob. `run_fp_client` is also run inside the zkVM
```

example/preloader/src/main.rs
```diff
@@ -37,27 +71,24 @@ where
...
-    // form of ELF. It is important to pass it to witness before calling PreloadedEigenDABlobProvider::from. Because the
-    // verification is checked within the elf
-
     let oracle = Arc::new(CachingOracle::new(
         ORACLE_LRU_SIZE,
         oracle_client,
         hint_client,
     ));
+
+    // Run derivation for the first time to populate the witness data
```
samlaf, line 82: does it also populate the oracle? you mention that below

example/preloader/src/main.rs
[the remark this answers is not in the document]
samlaf: Oh wait is this example all only running natively on the host system, and not inside zkvm? That wasn't clear to me up until reading this. Should probably be mentioned as the first thing in the README of this section. Also should make the above paragraph more clear.

README.md
```diff
@@ -42,4 +42,13 @@ Then you can run
...
 ```
 
+### Running the example preloaded client
+More information please see [here](./exmaple/preloader/README.md)
+
+```bash
+# Before running the client, it will download the needed g1.point SRS file
+# and the rollup.json config file.
+just run-client-against-devnet 'native' 'hokulea-example-preloader'
+```
```
samlaf, line 52: feel like this should live in the examples README, not here.

example/preloader/src/main.rs
samlaf, line 1: should we add this example to CI?

bin/host/src/lib.rs
```diff
@@ -3,3 +3,34 @@ pub mod eigenda_blobs;
 pub mod cfg;
 
 pub mod handler;
+
+use tracing_subscriber::{filter::LevelFilter, prelude::*, EnvFilter};
+pub fn init_tracing_subscriber(verbosity_level: u8) -> anyhow::Result<(), anyhow::Error> {
```
samlaf, line 8: should remove the one in main.rs then right?

bin/host/src/main.rs
```diff
@@ -17,7 +17,7 @@ async fn main() -> anyhow::Result<()> {
     Ok(())
 }
 
-fn init_tracing_subscriber(verbosity_level: u8) -> anyhow::Result<(), anyhow::Error> {
+pub fn init_tracing_subscriber(verbosity_level: u8) -> anyhow::Result<(), anyhow::Error> {
```
samlaf, line 20: this function should be removed now that you've put it in lib.rs no?

crates/proof/src/eigenda_witness_provider.rs
```diff
@@ -0,0 +1,64 @@
...
+use hokulea_eigenda::{AltDACommitment, EigenDABlobProvider, EigenDAVersionedCert};
+use hokulea_proof::cert_validity::CertValidity;
+use hokulea_proof::eigenda_blob_witness::EigenDABlobWitnessData;
+use rust_kzg_bn254_primitives::blob::Blob;
+use std::sync::{Arc, Mutex};
+
+/// This is a wrapper around OracleEigenDAProvider, with
+/// additional functionalities to generate eigenda witness
+/// which is KZG proof on the FS point out of the blob itself.
+/// OracleEigenDAWitnessProvider is only inteneded to be used outside
```
samlaf, line 13:
```suggestion
/// OracleEigenDAWitnessProvider is only intended to be used outside
```

crates/proof/src/eigenda_witness_provider.rs
```diff
@@ -0,0 +1,64 @@
...
+/// FPVM or ZKVM. Its sole purpose is to generate KZG proof at the
+/// client side
+#[derive(Debug, Clone)]
+pub struct OracleEigenDAWitnessProvider<T: EigenDABlobProvider> {
```
samlaf, line 17: should we call it `OracleEigenDAKZGProofProvider` then? Or `KZGWitnessProvider`?

crates/proof/src/eigenda_witness_provider.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// Eigenda provider
+    pub provider: T,
+    /// Store witness data
+    pub witness: Arc<Mutex<EigenDABlobWitnessData>>,
```
samlaf, line 21: does this need to be an ArcMutex?

crates/proof/src/eigenda_witness_provider.rs
```diff
@@ -0,0 +1,64 @@
...
+impl<T: EigenDABlobProvider + Send> EigenDABlobProvider for OracleEigenDAWitnessProvider<T> {
+    type Error = T::Error;
+
+    /// This function populates
+    /// 1. eigenda cert
+    /// 2. eigenda blob
+    /// 3. kzg blob proof on the random FS point
+    /// 4. CertValidity with claimed_validity = true, and receipt = None
+    /// The receipt is intended to assert the eigenda cert passing the view call on chain
+    /// The receipt generation is not included, it is expected that it takes significantly more time
```
samlaf, line 34: not following. when is the receipt generated/needed?

crates/proof/src/eigenda_witness_provider.rs
```diff
@@ -0,0 +1,64 @@
...
+            EigenDAVersionedCert::V2(c) => c,
+        };
+
+        // only a single blob is returned from a cert
+        let blob = self.provider.get_blob(altda_commitment).await?;
+
+        // Compute kzg proof for the entire blob on a deterministic random point
+        let kzg_proof = match compute_kzg_proof(blob.data()) {
+            Ok(p) => p,
+            Err(e) => panic!("cannot generate a kzg proof: {}", e),
```
samlaf, line 48: why/how can this fail? do we have ways to handle this without panicing?

example/preloader/src/main.rs
```diff
@@ -20,13 +28,39 @@ use alloy_evm::{EvmFactory, FromRecoveredTx, FromTxWithEncoded};
...
+    let cfg = SingleChainHostWithEigenDA::try_parse()?;
+    init_tracing_subscriber(cfg.verbose)?;
+
+    let hint = BidirectionalChannel::new()?;
+    let preimage = BidirectionalChannel::new()?;
+
+    let server_task = cfg.start_server(hint.host, preimage.host).await?;
+    // Start the client program in a separate child process.
+
+    let client_task = task::spawn(run_derivation_with_witgen_and_preloader(
```
samlaf, line 42:
```suggestion
    
    // Spawn the client logic as a concurrent task
    let client_task = task::spawn(run_derivation_with_witgen_and_preloader(
```

samlaf: its not a separate process, its just a tokio task in the same process

example/preloader/src/main.rs
```diff
@@ -20,13 +28,39 @@ use alloy_evm::{EvmFactory, FromRecoveredTx, FromTxWithEncoded};
 use op_alloy_consensus::OpTxEnvelope;
 use op_revm::OpSpecId;
 
+#[tokio::main(flavor = "multi_thread")]
```
samlaf, line 31: do we really benefit from the multi_threaded runtime? All we have is client and server running on 2 tasks, and all client calls are blocking right? So I don't think they ever run in parallel anyways?

example/preloader/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+Use the preloading method to verify the eigenda blob by converting witness data.
+
+## Witgen client
+
+Witgen client is a variant of the default fault proof client (that runs derivation pipeline and execution to check if output state matches). Running the default fault proof client produce an data oracle in the form of Key-Value map. There is data organization, all the keys are 32 bytes hash digest. A witgen client is a wrapper around the default client, that not only returns the oracle, but also an organized data structure called EigenDABlobWitnessData.
+
+### EigenDABlobWitnessData
+
+EigenDABlobWitnessData contains the EigenDA certificates (aka eigenda cert). The certificate is stored in an append only vector. For each cert, there is a corresponding eigenda blob, a kzg proof (which shows the kzg commitment relation between the blob and the eigenda cert) and a cert validity zk proof that contains the kzg commitment.
```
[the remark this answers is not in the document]
bxue-l2: that attests the validity of the da cert.

example/preloader/src/main.rs
bxue-l2: I guess so, kona is using it, https://github.com/op-rs/kona/blob/892743e1c02b2106132ee0fd0464409140dbcc28/bin/host/src/bin/host.rs#L48

crates/proof/src/eigenda_witness_provider.rs
bxue-l2: I don't think there is legitimate case it should fail, because if you are generating a kzg proof, unless blob is empty, it should always return somethign

crates/proof/src/eigenda_witness_provider.rs
bxue-l2: should have removed this file

crates/proof/src/eigenda_witness_provider.rs
bxue-l2: It is easier to be able to clone things

example/preloader/README.md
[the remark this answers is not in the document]
samlaf: I'm saying is the last 4 words of this paragraph correct? Or should it be "contains the kzg proof"?

example/preloader/src/main.rs
samlaf: doesn't mean they have thought about it, but anyways its just a nit.

example/preloader/README.md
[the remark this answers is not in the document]
bxue-l2: it was not correct, it should be "that attests the validity of the da cert."

crates/proof/src/eigenda_witness_provider.rs
bxue-l2: should have removed this file

example/preloader/README.md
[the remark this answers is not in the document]
bxue-l2: 7563ff22400e64158bbd3a1a74f777537d825805

crates/proof/src/eigenda_witness_provider.rs
bxue-l2: 7563ff22400e64158bbd3a1a74f777537d825805

crates/proof/src/eigenda_witness_provider.rs
bxue-l2: 7563ff22400e64158bbd3a1a74f777537d825805

PR #92, opened 2025-05-12
Refactor  use new kona trait avoid custom oracle and bump kona to 1.0.1

bxue-l2:
Incorporate Rami's [PR](https://github.com/op-rs/kona/pull/1655) on refactoring the derivation pipeline. 

Such that it is possible to remove the custom oracle pipeline file


Cargo.toml
```diff
@@ -13,25 +13,23 @@ codegen-units = 1
...
-kona-std-fpvm = { version = "0.2.0", git = "https://github.com/op-rs/kona", tag = "kona-client/v1.0.0", default-features = false }
-kona-std-fpvm-proc = { version = "0.2.0", git = "https://github.com/op-rs/kona", tag = "kona-client/v1.0.0", default-features = false }
-kona-providers-alloy = { version = "0.3.0", git = "https://github.com/op-rs/kona", tag = "kona-client/v1.0.0", default-features = false }
+kona-client = { git = "https://github.com/op-rs/kona", tag = "kona-client/v1.0.1", default-features = false }
+kona-host = { git = "https://github.com/op-rs/kona", tag = "kona-client/v1.0.1", default-features = false }
+kona-derive = { version = "0.3.0", git = "https://github.com/op-rs/kona", tag = "kona-client/v1.0.1", default-features = false }
+kona-driver = { version = "0.3.0", git = "https://github.com/op-rs/kona", tag = "kona-client/v1.0.1", default-features = false }
+kona-executor = { version = "0.3.0", git = "https://github.com/op-rs/kona", tag = "kona-client/v1.0.1", default-features = false }
+kona-mpt = { version = "0.2.0", git = "https://github.com/op-rs/kona", tag = "kona-client/v1.0.1", default-features = false }
```
samlaf, line 21: so they released v1.0.1 on github only but didn't push to crates.io (which is still on 0.3.0)? might be good to ask them to push to crates.io?

crates/client/src/fp_client.rs
```diff
@@ -5,19 +5,21 @@ use core::fmt::Debug;
 use alloy_consensus::Sealed;
 use tracing::{error, info};
 
-use hokulea_eigenda::EigenDABlobProvider;
-use hokulea_proof::pipeline::OraclePipeline;
+use hokulea_eigenda::{EigenDABlobProvider, EigenDABlobSource, EigenDADataSource};
+//use hokulea_proof::pipeline::OraclePipeline;
```
samlaf, line 9: delete

samlaf (approved):

Cargo.toml
bxue-l2: it is a bug fix

samlaf: Nvm 1.0.1, what about 1.0.0, why is that not on crates.io?

PR #96, opened 2025-05-14
Add justfile option to run against sepolia

bxue-l2:
This PR adds procedures to run against sepolia testnet


justfile
```diff
@@ -88,7 +88,7 @@ get-l2-finalize-block-number enclave='eigenda-devnet' chain_id='2151908':
 
 # Run the client program natively with the host program attached, against the op-devnet.
 [group('local-env')]
-run-client-against-devnet native_or_asterisc='native' env_file='.devnet.env' features='' mock_mode='' verbosity='' block_number='' rollup_config_path='rollup.json' enclave='eigenda-devnet' chain_id='2151908': (download-srs) (_download-rollup-config-from-kurtosis) (_kurtosis_wait_for_first_l2_finalized_block)
+run-client-against-devnet native_or_asterisc='native' verbosity='' env_file='.devnet.env' block_number='' rollup_config_path='rollup.json' enclave='eigenda-devnet' chain_id='2151908': (download-srs) (_download-rollup-config-from-kurtosis) (_kurtosis_wait_for_first_l2_finalized_block)
```
samlaf, line 91: rollup_config_path no longer used? there's probably other variables that can be cleaned up.

README.md
```diff
@@ -52,5 +52,17 @@ More information at [./example/preloader/README.md](./example/preloader/README.m
...
+## Run Against Sepolia
+
+You will need to run an instance of [eigenda-proxy](https://github.com/Layr-Labs/eigenda-proxy) with V2 support. Then populate the `.sepolia.env` file, see a template at `.example.env`.
+
+```bash
+# To download a `sepolia.rollup.json` from a rollup consensus node, you can use the command
+cast rpc "optimism_rollupConfig" --rpc-url $ROLLUP_NODE_RPC | jq > sepolia.rollup.json
+# Before running the client, it will download the needed g1.point SRS file
+# and the rollup.json config file. Temporary env variables are stored at
+# .devnet.env and .devnet.env.run
```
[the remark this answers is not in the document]
bxue-l2: .devnet.run.env needs more a bit more time, since you have to insert something in between, what if there are multiple "."

so go for .run.devnet.env

justfile
bxue-l2: actually it is used, will add it back

samlaf (approved):

PR #98, opened 2025-05-15
Refactor  common preloader

bxue-l2:
This PR did three small things

- reduce empty feature flags - it reduces number of run from 30 to 21
- remove unused l1_head_bn from cert_validity for zk verification, and create canoe input struct
- refactor common-preloader
  - run_within_zkvm has a dedicated crate named as zkvm_verification, which provides an easy interface for any zkVM to verify EigenDABlobwitness, the interface takes a witness and returns a preloadedEigenDaBlobPRovdier.


canoe/provider/src/lib.rs
```diff
@@ -1,17 +1,24 @@
...
 use anyhow::Result;
 use async_trait::async_trait;
-use hokulea_proof::cert_validity::CertValidity;
 use serde::{Deserialize, Serialize};
 
+pub struct CanoeInput {
+    /// eigenda cert
+    pub eigenda_cert: eigenda_v2_struct::EigenDAV2Cert,
+    /// the claim about if the cert is valid
+    pub claimed_validity: bool,
```
samlaf, line 10: can you add more info as to why this is needed again.
Keep forgetting, why can't we just prove the validity of the cert inside the zkvm? We need to do something with the validity result before we are ready to prove the cert valid right? or something like that?

canoe/provider/src/lib.rs
```diff
@@ -1,17 +1,24 @@
...
+    /// block hash where view call anchored at, l1_head comes from kona_cfg
```
samlaf, line 11: which view call? verifyCert view call? say that

canoe/provider/src/lib.rs
```diff
@@ -1,17 +1,24 @@
...
 
+pub struct CanoeInput {
+    /// eigenda cert
+    pub eigenda_cert: eigenda_v2_struct::EigenDAV2Cert,
+    /// the claim about if the cert is valid
+    pub claimed_validity: bool,
+    /// block hash where view call anchored at, l1_head comes from kona_cfg    
+    pub l1_head_block_hash: B256,
+    /// block number corresponding to the hash above. This is checked against l1_head_block_hash in the zk view proof
+    pub l1_head_block_number: u64,
```
samlaf, line 14:
```suggestion
    /// Block number corresponding to l1_head_block_hash.
    /// Their correspondence is checked in the zk view proof.
    pub l1_head_block_number: u64,
```

samlaf: better to not use referential statements like "hash above" because structs do change and their ordering could be flipped eventually by some refactor.

canoe/steel/apps/src/apps.rs
```diff
@@ -38,12 +37,13 @@ impl CanoeProvider for CanoeSteelProvider {
 
     async fn create_cert_validity_proof(
         &self,
-        eigenda_cert: eigenda_v2_struct::EigenDAV2Cert,
-        cert_validity: CertValidity,
+        canoe_input: CanoeInput,
+        //eigenda_cert: eigenda_v2_struct::EigenDAV2Cert,
+        //cert_validity: CertValidity,
```
samlaf, line 42:
```suggestion
```

crates/witgen/src/cert_validity_provider.rs
```diff
@@ -0,0 +1,32 @@
+use alloy_primitives::{BlockNumber, B256};
+use canoe_provider::{CanoeInput, CanoeProvider};
+use hokulea_proof::eigenda_blob_witness::EigenDABlobWitnessData;
+
+/// Populate canoe proof into cert validity. It is placed inside the normal crate because canoe
+/// currently assumes serde_json for serialization.
+pub async fn populate_cert_validity_to_witness(
+    witness: &mut EigenDABlobWitnessData,
```
samlaf, line 8: what are the assumptions on witness? Is it already half populated? Or is it just initialized and all fields are initialized by this function? If so, why not just return it as a value instead of passing a reference?

Typically when passing a &mut ref its best to describe why, and what are the assumptions around its use.

crates/zkvm-verification/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+) -> Result<PreloadedEigenDABlobProvider, FaultProofProgramError>
+where
+    O: CommsClient + FlushableCache + Send + Sync + Debug,
+{
+    let boot_info = BootInfo::load(oracle.as_ref()).await?;
+
+    // it is critical that some field of the witness is populated inside the zkVM using known truth within the zkVM
+    let num_cert = witness.validity.len();
+    for i in 0..num_cert {
+        witness.validity[i].l1_head_block_hash = boot_info.l1_head;
```
samlaf, line 33: seems like you just need l1_head from boot_info. Why do you pass in a full oracle here instead of just the l1_head directly?

example/preloader/src/main.rs
```diff
@@ -58,3 +86,136 @@ async fn main() -> anyhow::Result<()> {
...
+        evm_factory,
+    )
+    .await?;
+
+    let wit = core::mem::take(eigenda_blobs_witness.lock().unwrap().deref_mut());
+
+    Ok(wit)
+}
+
+// By this time,both Oracle and EigenDABlobWitnessData are generated by some party thxaxt runs run_witgen_client.
```
samlaf, line 196:
```suggestion
// By this time, both Oracle and EigenDABlobWitnessData are generated by some party that runs run_witgen_client.
```

samlaf (approved):

crates/zkvm-verification/src/lib.rs
bxue-l2: I am going to pass chain_id and sequence_window in the future, probably a good reason to add a struct

canoe/provider/src/lib.rs
bxue-l2: This flag is used when proxy tells the host that this cert is wrong. What canoe proof provides is that from EVM's perspective the result of this call. We need to make sure the host does not lie to us. Once it is checked. We are safe to ignore that cert.

There is potential optimization we can do, let chat's later

bxue-l2: Oh, your question is something else.

Yes, it is good to check what we are proving equals to what we expect. No need to waste effort.
Add comments for it

crates/zkvm-verification/src/lib.rs
bxue-l2: but actually, this is intentional, because the bootInfo in the oracle is supposed to already been trusted, so it would have to be verified by kona framework already @samlaf

PR #101, opened 2025-05-16
Chore  add risc0 sp1 installation in readme

bxue-l2:


README.md
```diff
@@ -39,6 +39,13 @@ Then you can run
 just run-client-against-devnet 'asterisc'
 ```
 
+### Risc0 and Sp1 toolchain installation
+
+Hokulea takes advantage of validity proof to attest DA certificates without making rpc call to execution client. The proof generation requires
```
samlaf, line 44:
```suggestion
Hokulea uses zk proofs to attest the validity of DA certificates without making rpc call to execution client. The proof generation requires 
```

samlaf: "without making rpc call to execution client"
That's not really the main goal right? main goal is to prove it onchain. Think the offchain benefit of not needing to run a full node (execution client) is not that important and shouldnt be emphasized here

README.md
```diff
@@ -39,6 +39,13 @@ Then you can run
 just run-client-against-devnet 'asterisc'
 ```
 
+### Risc0 and Sp1 toolchain installation
+
+Hokulea takes advantage of validity proof to attest DA certificates without making rpc call to execution client. The proof generation requires 
+compiling rust code into a ELF file runnable within zkVM. The canoe crate in Hokulea is dedicated for such purpose.
+Canoe currently supports two backends: Steel(Risc0) and Sp1-contract-call(Succinct).
+Steel requires installing Risc0 toolchain, see [rzup](https://dev.risczero.com/api/zkvm/install). Sp1-contract-call(sp1-cc) requires installing Sp1 toolchain, see [sp1up](https://docs.succinct.xyz/docs/sp1/getting-started/install). Try running or building binary with those features will fail.
```
samlaf, line 47:
```suggestion
Canoe currently supports two backends:
1. Steel(Risc0)
   - Gated behind the hokulea-proof [steel](https://github.com/Layr-Labs/hokulea/blob/3599bbeb855156164643a2a56c4f92de0cf7b7cf/crates/proof/Cargo.toml#L44) feature.
   - Requires installing the Risc0 toolchain, see [rzup](https://dev.risczero.com/api/zkvm/install).
2. Sp1-contract-call(Succinct)
   - Gated behind the hokulea-proof [sp1-cc](https://github.com/Layr-Labs/hokulea/blob/3599bbeb855156164643a2a56c4f92de0cf7b7cf/crates/proof/Cargo.toml#L45) feature.
   - Requires installing Sp1 toolchain, see [sp1up](https://docs.succinct.xyz/docs/sp1/getting-started/install).

Trying to build the hokulea client binary with either zkvm backend feature will fail if the respective toolchain is not installed.
```

README.md
bxue-l2: I kind of thinking zk proof means zero knowledge proof, which has information obfuscation process, which we don't

bxue-l2: the goal is to verified onchain or in zkvm. If zkvm can make rpc call, we won't have to spend so much time, like magic TEE making external call

bxue-l2: but I will just remove this part, since it is not that relevant for installation

bxue-l2: fcb63301bdea2993ae6edc8756a521ae4da6b962

samlaf (approved):

PR #122, opened 2025-05-28
fix: update new kurtosis syntax

bxue-l2:
The latest commit of Kurtosis breaks DA server

Submitted this PR(https://github.com/ethpandaops/optimism-package/pull/290) to their repo for fix, if approved, we should merge this PR


bxue-l2:
merge now, given kurtosis broken anyway. And it makes it easier to use local package

PR #133, opened 2025-06-07
Fix  hokulea work with v3 cert

bxue-l2:
This PR
- replace EigenDACertV2 from important structs with AltdaCommitment, allowing the system to hanlde V3 cert
  - validate_cert_receipt from CanoeVerifier trait
  - EigenDABlobWitnessData
  - PreloadedEigenDABlobProvider
  - CanoeInput
- add canoe support to V3
  - creating new v3 interface, abi binding including statusCode
  - refactor stdin to zkVM, now passing canoe_input. Remove many boilerplate
- add router address based on chain_id and Altdacommitment


bxue-l2:
Temporary removing kurtosis cache, because it is running out of disk. See https://github.com/Layr-Labs/hokulea/actions/runs/15515794758/job/43682504312?pr=135 for fix

.github/workflows/kurtosis_devnet.yaml
```diff
@@ -19,12 +19,6 @@ jobs:
           version: ${{ env.MISE_VERSION }}
           experimental: true
       - run: just run-kurtosis-devnet
-      - uses: Swatinem/rust-cache@v2 # cache for `just build-native`
-      # We build the client here because `just run-client-native-against-devnet`
-      # first polls op-geth for a first finalized block, which takes a few minutes,
-      # and only after that runs `cargo run` which then builds the client, and takes another
-      # few minutes. Better to build here while waiting for first l2 block to finalize.
-      - run: RISC0_SKIP_BUILD=1 SP1_SKIP_PROGRAM_BUILD=true just build-native
```
samlaf, line 27: why are you removing this? The failing CI you linked to just seems to be related to a crate name having changed?
image

canoe/bindings/src/lib.rs
```diff
@@ -92,8 +92,24 @@ sol! {
         uint32[][] nonSignerStakeIndices;
     }
 
+    struct EigenDACertV3 {
+        BatchHeaderV2 batchHeaderV2;
+        BlobInclusionInfo blobInclusionInfo;
+        NonSignerStakesAndSignature nonSignerStakesAndSignature;
+        bytes signedQuorumNumbers;
```
samlaf, line 99: `signedQuorumNumbers` is such a bad name. Is that what its called in the contract and golang code? Is this the signed quorum numbers on the batch? How about we rename to include `batch` in the name here?

canoe/provider/src/lib.rs
```diff
@@ -43,3 +47,32 @@ impl CanoeProvider for CanoeNoOpProvider {
...
     }
 }
+
+/// Call respecting solidity interface
+/// V2 is deprecated once router is released
+pub enum CertVerifierCall {
+    /// V2 calldata
+    V2(IEigenDACertVerifier::verifyDACertV2ForZKProofCall),
+    /// Router calldata for V3 and beyond
+    Router(IEigenDACertVerifierRouter::checkDACertCall),
```
samlaf, line 57: what if there is no router but the V3CertVerifier is used directly?

canoe/provider/src/lib.rs
```diff
@@ -43,3 +47,32 @@ impl CanoeProvider for CanoeNoOpProvider {
...
+}
+
+/// convert eigenda cert type into its solidity type that works with solidity cert verifier interface
+pub fn build_call(altda_commitment: &AltDACommitment) -> CertVerifierCall {
```
samlaf, line 61: change this to a static method on CertVerifierCall:

```
impl CertVerifierCall {
    pub fn new(altda_commitment: &AltDACommitment) -> Self {
        match &altda_commitment.versioned_cert {
            // ...
        }
    }
}
```

canoe/provider/src/lib.rs
samlaf: ah we're only calling the checkDACert() on the router, which has same exact interface as v3CertVerifier right? So even if its just a CertVerifierV3 itll still work? Prob worth commenting this

canoe/sp1-cc/client/src/main.rs
```diff
@@ -16,45 +15,31 @@ pub fn main() {
...
-
-    let signed_quorum_numbers = Bytes::abi_decode(&signed_quorum_numbers_abi)
-        .expect("deserialize signed_quorum_numbers_abi");
-
-    let mock_call = IEigenDACertVerifier::verifyDACertV2ForZKProofCall {
-        batchHeader: batch_header,
-        blobInclusionInfo: blob_inclusion_info.clone(),
-        nonSignerStakesAndSignature: non_signer_stakes_and_signature,
-        signedQuorumNumbers: signed_quorum_numbers,
+    // TODO, are there no better way to reduce this replicated code.
```
samlaf, line 24:
```suggestion
    // TODO, are there no better way to reduce this duplicate code.
```

canoe/sp1-cc/client/src/main.rs
```diff
@@ -16,45 +15,31 @@ pub fn main() {
...
+    // known constraint, new_call takes SolCall trait, which is Sized so not dyn trait
+    // also impl SolCall seems too much
```
samlaf, line 26: not sure what you are talking about here, is this comment really needed or is it stream of consciousness?

samlaf: what does "seems too much" mean?

canoe/sp1-cc/client/src/main.rs
```diff
@@ -63,14 +48,12 @@ pub fn main() {
...
-        input: buffer.into(),
+        input: rlp_bytes.into(),
         blockhash: public_vals.anchorHash,
         output: returns,
         l1ChainId: chain_sepc.chain.id(),
     };
 
     // Commit the abi-encoded output.
-    // We can't use ContractPublicValues because sp1_cc_client_executor currently has deps issue.
-    // Instead we define custom struct to commit
```
samlaf, line 74: was this fixed on succinct side?

canoe/steel/methods/contracts/Elf.sol
```diff
@@ -19,6 +19,6 @@
 pragma solidity ^0.8.20;
 
 library Elf {
-    string public constant V2CERT_VERIFICATION_PATH =
-        "/Users/bxue/Documents/eigenda-integration/hokulea/target/riscv-guest/canoe-steel-methods/v2cert-verification-guests/riscv32im-risc0-zkvm-elf/release/v2cert_verification.bin";
+    string public constant CERT_VERIFICATION_PATH =
+        "/Users/bxue/Documents/eigenda-integration/hokulea/target/riscv-guest/canoe-steel-methods/cert-verification-guests/riscv32im-risc0-zkvm-elf/release/cert_verification.bin";
```
samlaf, line 23: don't check in a path that only works on your computer

canoe/bindings/src/lib.rs
```diff
@@ -92,8 +92,24 @@ sol! {
...
+    }
+
+    #[derive(PartialEq)]
+    enum StatusCode {
```
samlaf, line 103: these are CertVerifierV3 error codes right? Like V4 could have new error codes I think? Prob should name `StatusCodeCVV3` or something?
Also what about generating these bindings using alloy? Is that too hard/annoying?

canoe/steel/apps/src/apps.rs
```diff
@@ -105,19 +93,16 @@ impl CanoeProvider for CanoeSteelProvider {
             let env = ExecutorEnv::builder()
                 .write(&evm_input)?
                 .write(&verifier_address)?
-                .write(&batch_header_abi)?
-                .write(&non_signer_abi)?
-                .write(&blob_inclusion_abi)?
-                .write(&signed_quorum_numbers_abi)?
-                .write(&canoe_input.l1_chain_id)?
+                .write(&canoe_input)?
+                //.write(&canoe_input.l1_chain_id)?
```
samlaf, line 97:
```suggestion
```

crates/eigenda-cert/src/altda_commitment.rs
```diff
@@ -147,10 +155,44 @@ impl AltDACommitment {
...
+                    .blob_header
+                    .commitment
+                    .commitment
+                    .y,
+            ),
+            EigenDAVersionedCert::V3(c) => (
+                c.blob_inclusion_info
+                    .blob_certificate
+                    .blob_header
+                    .commitment
```
samlaf, line 179: might be cleaner to define an interface and have each versionedCert implement it? Another instance of https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html

samlaf: Not sure if it works with your current code, or how hard it is to refactor though

bxue-l2:
> https://github.com/Layr-Labs/hokulea/actions/runs/15515794758/job/43682504312?pr=135

You are looking at the pr, that I tried to fix but forgot to add --bin. The real issue can be found at the PR I linked at the top.

canoe/steel/methods/contracts/Elf.sol
bxue-l2: this is automatically generated by risc0, if you download to your computer the path will be overwritten

crates/eigenda-cert/src/altda_commitment.rs
bxue-l2: this is the type polymorphism you talked about.

crates/proof/src/canoe_verifier/mod.rs
```diff
@@ -11,46 +11,65 @@ use alloc::vec::Vec;
...
+        EigenDAVersionedCert::V3(_) => cert_verifier_router_address(chain_id),
+    }
+}
+
+pub fn cert_verifier_router_address(chain_id: u64) -> Address {
+    // this is kurtosis devnet
+    match chain_id {
+        // mainnet
+        1 => address!("0x61692e93b6B045c444e942A91EcD1527F23A3FB7"),
+        // Sepolia V2 cert verifier address
```
bxue-l2, line 58: V3 cert

canoe/provider/src/lib.rs
bxue-l2: we hardcoded the router address in the client, see func `cert_verifier_router_address`

canoe/provider/src/lib.rs
bxue-l2: 3f1955bb4efa0d167ffb8567e8b39042bc398a91

canoe/sp1-cc/client/src/main.rs
bxue-l2: 3f1955bb4efa0d167ffb8567e8b39042bc398a91

canoe/bindings/src/lib.rs
bxue-l2: add comments on what it is

canoe/steel/methods/contracts/Elf.sol
bxue-l2: put in gitignore

canoe/bindings/src/lib.rs
bxue-l2: d2857d38cf50b52fbcd7b29dd30d834c6e1b14b7

samlaf (approved):

PR #143, opened 2025-06-14
refactor canoe proof gen and replace error types

### LINKED ISSUE #142: [refactor] wrap canoe proof generation into a single function
Currently, after running the OracleEigenDAWitnessProvider, the user (zkVM host) must copy almost 30 lines code to generate and overwrite the canoe_proof, see https://github.com/Layr-Labs/hokulea/blob/master/example/preloader/src/main.rs#L145-L178

We should create a function that accepts BootInfo and CanoeProvider and return the Canoe Proof, (i.e the journal)

bxue-l2:
This PR refactor canoe_proof_gen from the preloader example into a helper function.

It is asked from op-succinct to reduce the size of code, their side has to copy in order to generate canoe proof. Although it is not a blocker at the moment.

This PR also changes the error type, FaultProofProgramError, in various places. The change makes the error type more accurate. Also, example should have used anyhow

Fix #142
Fix DAINT-572

### DIFF
--- a/crates/witgen/src/canoe_witness_provider.rs
+++ b/crates/witgen/src/canoe_witness_provider.rs
@@ -23,7 +23,8 @@ where
     let header_rlp = oracle
         .get(PreimageKey::new_keccak256(*boot_info.l1_head))
         .await?;
-    let l1_head_header = Header::decode(&mut header_rlp.as_slice())?;
+    let l1_head_header = Header::decode(&mut header_rlp.as_slice())
+        .map_err(|_| anyhow::Error::msg("cannot rlp decode header in canoe proof generation"))?;
     let l1_chain_id = boot_info.rollup_config.l1_chain_id;
 
     let mut wit = witness.clone();
--- a/crates/witgen/src/canoe_witness_provider.rs
+++ b/crates/witgen/src/canoe_witness_provider.rs
@@ -1,0 +1,55 @@
+use alloy_consensus::Header;
+use alloy_rlp::Decodable;
+use canoe_provider::{CanoeInput, CanoeProvider};
+use core::fmt::Debug;
+use hokulea_proof::eigenda_blob_witness::EigenDABlobWitnessData;
+use kona_preimage::{CommsClient, PreimageKey};
+use kona_proof::{BootInfo, FlushableCache};
+use std::sync::Arc;
+
+/// A helper function to create canoe proof by the provided canoe provider.
+/// The function relies on data stored in the oracle, for l1_head, l1_head_number
+/// and chain_id.
+pub async fn from_boot_info_to_canoe_proof<P, O>(
+    boot_info: &BootInfo,
+    witness: &EigenDABlobWitnessData,
+    oracle: Arc<O>,
+    canoe_provider: P,
+) -> anyhow::Result<Vec<P::Receipt>>
+where
+    P: CanoeProvider,
+    O: CommsClient + FlushableCache + Send + Sync + Debug,
+{
+    let header_rlp = oracle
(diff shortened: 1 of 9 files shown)


crates/witgen/src/canoe_witness_provider.rs
```diff
@@ -0,0 +1,55 @@
...
+    canoe_provider: P,
+) -> anyhow::Result<Vec<P::Receipt>>
+where
+    P: CanoeProvider,
+    O: CommsClient + FlushableCache + Send + Sync + Debug,
+{
+    let header_rlp = oracle
+        .get(PreimageKey::new_keccak256(*boot_info.l1_head))
+        .await
+        .expect("get l1 header based on l1 head");
```
samlaf, line 26: your function returns an anyhow::Result but you use `expect()` everywhere instead of propagating errors. Either return a Result and propagate errors with `?`, or instead use `expect()` everywhere and return a `Vec<P::Receipt>` directly (not wrapped in Result)

crates/witgen/src/canoe_witness_provider.rs
```diff
@@ -0,0 +1,55 @@
...
+use alloy_rlp::Decodable;
+use canoe_provider::{CanoeInput, CanoeProvider};
+use core::fmt::Debug;
+use hokulea_proof::eigenda_blob_witness::EigenDABlobWitnessData;
+use kona_preimage::{CommsClient, PreimageKey};
+use kona_proof::{BootInfo, FlushableCache};
+use std::sync::Arc;
+
+/// A helper function to create canoe proof by the provided canoe provider.
+/// The function relies on data stored in the oracle, for l1_head, l1_head_number
```
samlaf, line 11: I don't see l1_head_number being used anywhere.

crates/witgen/src/canoe_witness_provider.rs
```diff
@@ -0,0 +1,55 @@
...
+    O: CommsClient + FlushableCache + Send + Sync + Debug,
+{
+    let header_rlp = oracle
+        .get(PreimageKey::new_keccak256(*boot_info.l1_head))
+        .await
+        .expect("get l1 header based on l1 head");
+    let l1_head_header = Header::decode(&mut header_rlp.as_slice()).expect("rlp decode l1 header");
+    let l1_chain_id = boot_info.rollup_config.l1_chain_id;
+
+    let mut wit = witness.clone();
```
samlaf, line 30: is this expensive in terms of zkvm ops? If so, can we consider changing the function signature to pass a mut witness instead?

example/preloader/src/main.rs
```diff
@@ -144,36 +142,24 @@ where
...
+    .await?;
+
+    // populate canoe proof for this example, in general canoe_proofs are used differently depending on
+    // where it is verified
+    // for verification within zkVM, canoe_proof should be passed in to zkVM via its stdin by a special
+    // function depending on zkVM framework. More see CanoeVerifier
+    // For Sp1cc, use CanoeSp1CCReducedProofProvider to produce proof that is verifiable within zkVM
+    // For Steel, use CanoeSteelProvider to generate such proof
+    // For verification in non zkVM context, canoe_proofs can be passed as part of serialized bytes
+    // along with other
```
samlaf, line 161: why don't you also pass them in via stdin to make all context behave the same? wouldnt that be simpler?

samlaf (approved):

example/preloader/src/main.rs
bxue-l2: for those special pass in, like add_assumption, the proof can only be received in the zkVM mode. Which uses zkvm specific library. I dont think you can do the testing in the normal mode

PR #155, opened 2025-08-14
Fix: canoe sp1-cc ELF compilation with new sp1v5 and miscellaneous fixes

bxue-l2:
This PR addresses several issues

- bumped sp1 to version 5
- increased the number of SRS to load to 16MiB worth of data
- remove auto compile to build ELF for sp1cc guest
- update ELF file and new vKey for sp1cc

Notes

- We need to remove the auto ELF build.rs because we currently the build.rs file doesn't pin the version of sp1 toolchain. As the result, when another project imports the canoe sp1-cc crate, it will automatically compile a new ELF file using some unknown toolchain, which inevitably is inconsistent with the vKey we hard-coded in the binary.


canoe/sp1-cc/client/src/main.rs
```diff
@@ -5,38 +5,39 @@ use alloy_primitives::Address;
...
-    // This step also validates all of the storage against state root provided by the host
-    let executor = ClientExecutor::new(&state_sketch).unwrap();
-
-    let chain_sepc: ChainSpec = executor
-        .genesis
-        .try_into()
-        .expect("convert sp1 genesis into chain spec");
+    // l1_chain_id is committed to the journal that would be compared aginast the journal generated by
+    // the hokulea program. Chain ID is checked implicitly
+    let l1_chain_id = executor.chain_spec.chain().id();
```
samlaf, line 40: dont you need to assert that this is == l1_chain_id_from_canoe_input above?

canoe/sp1-cc/client/src/main.rs
```diff
@@ -5,38 +5,39 @@ use alloy_primitives::Address;
...
     let canoe_inputs = sp1_zkvm::io::read::<Vec<CanoeInput>>();
 
     // ensure all canoe_proof uses identical l1 chain id and l1 head block number
     assert!(!canoe_inputs.is_empty());
-    let l1_chain_id = canoe_inputs[0].l1_chain_id;
+
+    // Initialize the client executor with the state sketch.
+    // This step also validates all of the storage against state root provided by the host
+    let executor =
+        ClientExecutor::eth(&state_sketch).expect("should be able to initialize client executor");
```
samlaf, line 26: why move this up here if its only used below?

crates/compute-proof/src/kzg_proof.rs
```diff
@@ -17,7 +17,7 @@ use rust_kzg_bn254_prover::srs::SRS;
 pub fn compute_kzg_proof(blob: &[u8]) -> Result<Bytes, KzgError> {
     let srs_file_path = "resources/g1.point";
     // In the future, it might make sense to let the proxy to return kzg proof, instead of local computation
-    let srs = SRS::new(srs_file_path, 268435456, 1024)
+    let srs = SRS::new(srs_file_path, 268435456, 65536)
```
samlaf, line 20: Isnt this only 64KiSymbols * 32 bytes/symbol = 2MiB?

crates/proof/src/canoe_verifier/sp1_cc.rs
```diff
@@ -2,15 +2,29 @@ use crate::canoe_verifier::errors::HokuleaCanoeVerificationError;
...
-pub const VKEYHEXSTRING: &str = "00692d1aac07e614ced12f4b16f50834490e735b89d2d3492b343337bb31d227";
+/// Any change to sp1-cc client including new sp1 toolchain produces a new ELF to be executed and proved by zkVM
+/// To generate the new ELF (a newer version than 5.2.1 toolchain tag is also fine)
+/// ``` bash
+/// cd canoe/sp1-cc/client
+/// cargo prove build --output-directory ../elf --elf-name canoe-sp1-cc-client --docker --tag v5.2.1
+/// ```
+///
+/// The verificaiton of the ELF must be hardcoded here which pins an exact version of ELF a prover can use
+/// Sp1 toolchain currently does not prove a way to generate such key. It has been raised to the sp1 team.
```
samlaf, line 17:
```suggestion
/// Sp1 toolchain currently does not provide a way to generate such key. It has been raised to the sp1 team.
```

samlaf (approved):

crates/compute-proof/src/kzg_proof.rs
bxue-l2: good catch

crates/proof/src/canoe_verifier/sp1_cc.rs
bxue-l2: 8c7bab0ad0f9b21635640c55703fc6e91db41c86

crates/compute-proof/src/kzg_proof.rs
bxue-l2: 8c7bab0ad0f9b21635640c55703fc6e91db41c86

canoe/sp1-cc/client/src/main.rs
bxue-l2: 8c7bab0ad0f9b21635640c55703fc6e91db41c86

PR #157, opened 2025-08-21
fix: implement encoded payload

bxue-l2:
Key Changes:
  - Renamed data struct: Changed from EigenDABlobData to EncodedPayload
  - Core implementation: translate proxy encoded payload impl to rust mainly by claude
  - Integration updates: Use a newer version of the rust kzg library

Kurtosis devnet parameter update 0d715d81dc12512fef9d23ac8b981ff5d4aa3c40:
- update proxy tag
- Changes in op-batcher parameter forces us to remove throttle parameter. See op change at [PR](https://github.com/ethereum-optimism/optimism/pull/17032). The stable kurtosis release uses the develop tag
Screenshot 2025-08-21 at 4 42 10 PM

We could have used a tag of op-batcher, but downside is that we could use a significantly old software in the future. So not tagging for now

 
Key callout
 - I was attempting to work on renaming in the beginning, but its scope would be very large. Renaming will appear in another PR
 - Before renaming, I will have a new PR to use the newer release of proxy that fetch the encoded payload

### DIFF
--- a/bin/host/src/handler.rs
+++ b/bin/host/src/handler.rs
@@ -5,7 +5,7 @@ use crate::status_code::{DerivationError, HostHandlerError, HTTP_RESPONSE_STATUS
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
 use eigenda_cert::AltDACommitment;
-use hokulea_eigenda::{EigenDABlobData, HokuleaPreimageError};
+use hokulea_eigenda::{EncodedPayload, HokuleaPreimageError};
 use hokulea_eigenda::{
     BYTES_PER_FIELD_ELEMENT, PAYLOAD_ENCODING_VERSION_0, RESERVED_EIGENDA_API_BYTE_FOR_RECENCY,
     RESERVED_EIGENDA_API_BYTE_FOR_VALIDITY, RESERVED_EIGENDA_API_BYTE_INDEX,
@@ -246,11 +246,11 @@ async fn store_blob_data(
     let mut kv_write_lock = kv.write().await;
     // Prepare blob data
     let blob_length_fe = altda_commitment.get_num_field_element();
-    let eigenda_blob = EigenDABlobData::encode(rollup_data.as_ref(), PAYLOAD_ENCODING_VERSION_0);
+    let encoded_payload = EncodedPayload::encode(rollup_data.as_ref(), PAYLOAD_ENCODING_VERSION_0);
 
     // Verify blob data is properly formatted
-    assert!(eigenda_blob.blob.len() % 32 == 0);
-    let fetch_num_element = (eigenda_blob.blob.len() / BYTES_PER_FIELD_ELEMENT) as u64;
+    assert!(encoded_payload.encoded_payload.len() % 32 == 0);
+    let fetch_num_element = (encoded_payload.encoded_payload.len() / BYTES_PER_FIELD_ELEMENT) as u64;
 
     // Store each field element
     let mut field_element_key = altda_commitment.digest_template();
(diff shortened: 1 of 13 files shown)


bin/host/src/handler.rs
```diff
@@ -246,12 +247,31 @@ async fn store_blob_data(
...
-
+    let encoded_payload = EncodedPayload::encode(rollup_data.as_ref(), PAYLOAD_ENCODING_VERSION_0);
     // Verify blob data is properly formatted
-    assert!(eigenda_blob.blob.len() % 32 == 0);
-    let fetch_num_element = (eigenda_blob.blob.len() / BYTES_PER_FIELD_ELEMENT) as u64;
+    assert!(encoded_payload.encoded_payload.len() % 32 == 0);
+
+    // Preliminary defense check against malicious eigenda proxy host
+    // Validate field elements (keeping existing field element validation for compatibility)
+    let encoded_payload_body = &encoded_payload.encoded_payload[ENCODED_PAYLOAD_HEADER_LEN_BYTES..];
```
samlaf, line 256: you don't check if encoded_payload has len 0 right? aka this could panic?

bin/host/src/handler.rs
```diff
@@ -246,12 +247,31 @@ async fn store_blob_data(
...
+    // verify there is an empty byte for every 31 bytes. This is a harder constraint than field element range check.
+    for chunk in encoded_payload_body.chunks_exact(BYTES_PER_FIELD_ELEMENT) {
+        // very conservative check on Field element range. It allows us to detect
+        // misbehaving at the host side when providing the field element. So we can stop early.
+        // the field element of on bn254 curve is some number less than 2^254
+        // that means both 255 and 254 th bits must be 0. out of conservation, we require the
+        // 253 bit to be 0. It aligns with our encoding scheme below that the first 8bits
+        // should be 0.
+        // Field elements are interpreted as big endian
+        if chunk[0] & 0b1110_0000 != 0 {
```
samlaf, line 266: why do you check the 3rd bit? What is this "out of conservation"?

bin/host/src/handler.rs
```diff
@@ -246,12 +247,31 @@ async fn store_blob_data(
...
+            return Err(anyhow!("invalid field element encoding"));
+        }
+        // we don't have the check that the first 8 bits are zero, because it is a more restrictive check, that might
+        // affect future payload encoding scheme
```
samlaf, line 270: move above with the other comment? kind of randomly placed here

crates/eigenda/src/eigenda_data.rs
```diff
@@ -90,11 +147,14 @@ impl EigenDABlobData {
     ///
     /// The length of (header + payload) by the encode function is always multiple of 32
     /// The eigenda proxy does not take such constraint.
+    ///
+    /// (ToDo) with new proxy release, it can return the encoded payload, the encode function can
```
samlaf, line 151:
```suggestion
    /// (ToDo) with new proxy release 2.2.1, it can return the encoded payload, the encode function can
```

samlaf (approved):

bin/host/src/handler.rs
```diff
@@ -246,12 +247,31 @@ async fn store_blob_data(
     let mut kv_write_lock = kv.write().await;
     // Prepare blob data
     let blob_length_fe = altda_commitment.get_num_field_element();
-    let eigenda_blob = EigenDABlobData::encode(rollup_data.as_ref(), PAYLOAD_ENCODING_VERSION_0);
-
+    let encoded_payload = EncodedPayload::encode(rollup_data.as_ref(), PAYLOAD_ENCODING_VERSION_0);
     // Verify blob data is properly formatted
-    assert!(eigenda_blob.blob.len() % 32 == 0);
-    let fetch_num_element = (eigenda_blob.blob.len() / BYTES_PER_FIELD_ELEMENT) as u64;
+    assert!(encoded_payload.encoded_payload.len() % 32 == 0);
```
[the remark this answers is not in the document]
bxue-l2: The code below can automatically pad to next power of 2. Similar to what go client does
https://github.com/Layr-Labs/hokulea/blob/07a649e3f7ed8d577e6febfbe2efc2923c3ac9b3/bin/host/src/handler.rs#L289

bin/host/src/handler.rs
bxue-l2: 21888242871839275222246405745257275088548364400416034343698204186575808495617 is 001xxx0xx

You can flip the last 0 to 1, and violate the FE range. So by setting it the third bit 0, it's guaranteed to be within the range. However, it doesn't cover the entire FE.

The bit check is pretty efficient

PR #167, opened 2025-09-26
add tests for eigenda blob derivation

bxue-l2:


crates/eigenda/src/eigenda_data.rs
```diff
@@ -1,3 +1,6 @@
+//! Contains Kona and EigenDA blob derivation pipeline. Typically rollup or
+//! proving stack use their own derivation pipeline with customization.
```
samlaf, line 2: do you mean that they would fork this module? or that they would use it as is but add modifications on top?

crates/eigenda/src/eigenda_data.rs
```diff
@@ -219,10 +232,175 @@ mod tests {
...
         assert!(result.is_err());
-        // after truncation, 33 - 32(header length) = 1
         assert_eq!(
             result.unwrap_err(),
-            EncodedPayloadDecodingError::InvalidLengthInEncodedPayloadBody(1).into()
+            EncodedPayloadDecodingError::InvalidLengthEncodedPayload(33).into()
+        );
+    }
+
+    #[test]
```
samlaf, line 241: idk how easy it is to do in rust but would be nice to have "table driven tests" the same way golang has. Makes it easy to see all the test cases being tested. Also would be curious if there's a way to assert that a test has exhaustively tested all error enum values.

samlaf (approved):

crates/eigenda/src/eigenda_data.rs
bxue-l2: They craft their own module using kona as basis

crates/eigenda/src/eigenda_data.rs
bxue-l2: bdf7d1ab9b9adb41c72b5132127959ad09514164

PR #168, opened 2025-09-26
tools: add tools to generate vkey

bxue-l2:
This PR adds a small tool to generate the vKey for Sp1CC ELF using sp1 sdk.

This PR also adds a justfile function to
- generate a new ELF with a known tag
- call the binary tool to get the vKey


samlaf (approved):
Didnd try to run.. probably should... but lazy. #test-in-prod
But seriously, this is the vkey that needs to be modified in the code right?
Prob good to add a comment as to that effect somewhere, both in the code above the vkey, as well as in the justfile after running the command. Something like
"make sure to paste this new vkey into your code in file xyz"

PR #169, opened 2025-09-27
test: add mock object and test for eigenda preimage

bxue-l2:
This PR creates a mock object implementing EigenDAPreimageProvider trait.

This PR also adds unit tests for eigenda_preimage.rs. 

This PR also renames some error and delete unused errors type in an enum.

### DIFF
--- a/crates/eigenda/src/errors.rs
+++ b/crates/eigenda/src/errors.rs
@@ -4,7 +4,7 @@ use alloc::string::{String, ToString};
 use eigenda_cert::AltDACommitmentParseError;
 
 /// Actionable hokulea error
-#[derive(Debug, thiserror::Error)]
+#[derive(Debug, thiserror::Error, PartialEq)]
 pub enum HokuleaErrorKind {
     /// for cert that has violated the rules in hokulea derivation
     #[error("Discard {0}")]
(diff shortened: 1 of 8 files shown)


crates/eigenda/src/errors.rs
```diff
@@ -17,8 +17,10 @@ pub enum HokuleaErrorKind {
     Temporary(String),
 }
 
-/// A list of Hokulea error purely out of data processing, and is decoupled from
-/// the error from error out of the preimage error
+/// A list of Hokulea application error purely out of data processing.
+/// They are at the same level as preimage error, except those error
+/// does not depends on the state returned from the preimage oracle.
+/// Those are higher level errors than provider error
```
samlaf, line 23: bit confused as to what those means here. stateless or preimage or both?

crates/eigenda/src/errors.rs
```diff
@@ -88,8 +84,9 @@ pub enum EncodedPayloadDecodingError {
 
 /// A list of Hokulea error derived from data from preimage oracle
 /// This error is intended for application logics, and it is separate from
-/// the more basic error type that deals with HokuleaOracleProviderError
-/// which hanldes communicates, response format error
+/// the more basic error type that deals with Provider error like
+/// HokuleaOracleProviderError which hanldes communicates and response format
```
samlaf, line 88:
```suggestion
/// HokuleaOracleProviderError which handles communicates and response format
```

samlaf: handles communicates?

samlaf (approved):

crates/eigenda/src/errors.rs
bxue-l2: Wanted to say, both HokuleaStatelessError and PreimageError defined in this file are application level error.

bxue-l2: is it better?

PR #173, opened 2025-10-01
chores: improve logging for 0 canoe proof generation

seolaoh:
I noticed a log message `producing 1 canoe proof for 0 DA certs` while testing EthDA proving, though canoe proof generation was skipped. This PR improves the logging for cases where canoe proof generation is skipped, making it clearer and based on DA cert availability.

### DIFF
--- a/crates/witgen/src/canoe_witness_provider.rs
+++ b/crates/witgen/src/canoe_witness_provider.rs
@@ -40,7 +40,11 @@ where
 
     let mut canoe_inputs = vec![];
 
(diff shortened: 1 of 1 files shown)


bxue-l2 (approved):

bxue-l2:
this is very strange that elf failed. I think it is due to [this](https://github.com/Layr-Labs/hokulea/blob/770874fe29e2dc54fa798e6249daf817ac91fac9/.github/workflows/elf.yml#L21)

But I have checked the ELF is unchanged

PR #174, opened 2025-10-02
feat: abstract address getter and cleanup canoe verifier

bxue-l2:
This PR abstract verifier address getter.

- both provider and verifier need to instantiate the CanoeVerifierAddressFetcher to learn about the address. Previously it was hardcoded
- after the change, a rollup can implement their own structure for specifying rollup deployed address
- We moved Address into CertValidity struct, because it opens the possibility for offchain routing, if a user does not want to use onchain router.

### DIFF
--- a/crates/proof/src/canoe_verifier/errors.rs
+++ b/crates/proof/src/canoe_verifier/errors.rs
@@ -6,10 +6,8 @@ use alloc::string::String;
 /// sp1 cannot take sp1-sdk as dependency which is needed for verification in non zkvm mode
 #[derive(Debug, thiserror::Error)]
 pub enum HokuleaCanoeVerificationError {
-    /// InconsistentPublicJournal
     #[error("Non zkvm environment: inconsistency between public journal proven by the zk proof and user supplied journal")]
     InconsistentPublicJournal,
-    /// MissingProof
     #[error("Non zkvm environment: proof is missing")]
     MissingProof,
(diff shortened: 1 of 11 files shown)


crates/proof/src/canoe_verifier/errors.rs
```diff
@@ -6,10 +6,10 @@ use alloc::string::String;
 /// sp1 cannot take sp1-sdk as dependency which is needed for verification in non zkvm mode
 #[derive(Debug, thiserror::Error)]
 pub enum HokuleaCanoeVerificationError {
-    /// Invalid Cert validity response
+    /// InconsistentPublicJournal
```
samlaf, line 9:
```suggestion
```

crates/proof/src/canoe_verifier/errors.rs
```diff
@@ -6,10 +6,10 @@ use alloc::string::String;
...
     #[error("Non zkvm environment: inconsistency between public journal proven by the zk proof and user supplied journal")]
     InconsistentPublicJournal,
-    /// Invalid Cert validity response
+    /// MissingProof
```
samlaf, line 12:
```suggestion
```

samlaf (approved):

crates/proof/src/canoe_verifier/errors.rs
bxue-l2: e3c7d330047af8548973184412aef30d38730d70

PR #175, opened 2025-10-02
ci: rm unnecessary ref in ci causing elf.yml fail

bxue-l2:
## Problem 
See this [Pr](https://github.com/Layr-Labs/hokulea/pull/173)

and CI [error](https://github.com/Layr-Labs/hokulea/actions/runs/18160673153/job/51760995467)

this cannot checkout branch from another PR, instead we should use the detached head from the PR which is the github default

### DIFF
--- a/.github/workflows/elf.yml
+++ b/.github/workflows/elf.yml
@@ -18,8 +18,6 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
-        with:
-          ref: ${{ github.event.pull_request.head.ref }}
       - name: Install SP1 toolchain
         run: |
           curl -L https://sp1.succinct.xyz | bash


samlaf (approved):
Believe thats the default branch it checks out anyways, so don't think this will change anything.

My guess for why that fails is b/c it was running in seolah's fork. Why do you need the ELF to be generated in CI for PRs? Is it to force PRs to regenerate the ELF?

bxue-l2:
> Believe thats the default branch it checks out anyways, so don't think this will change anything.
> 
> My guess for why that fails is b/c it was running in seolah's fork. Why do you need the ELF to be generated in CI for PRs? Is it to force PRs to regenerate the ELF?

yes, regenerate the elf and check it has not changed. Otherwise, we need update the ELF and vkey for sp1cc

PR #310, opened 2026-04-15
fix: relax alloy-consensus exact pins to semver ranges

fakedev9999:
## Summary
Relax `alloy-consensus = "=1.6.3"` and `op-alloy-consensus = "=0.23.1"` exact pins to semver ranges (`"1.6.3"` and `"0.23.1"`).

## Why
The exact pins prevent downstream consumers (op-succinct) from resolving alloy 1.7.x (required by kona). Semver ranges allow cargo to unify versions across the dependency tree.

## Changes
- `alloy-consensus`: `"=1.6.3"` → `"1.6.3"`
- `op-alloy-consensus`: `"=0.23.1"` → `"0.23.1"`

That's it — 2 characters removed.

## Request
Please merge and cut a release tag (e.g. `hokulea-client/v1.1.7`) so op-succinct can reference it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,13 +56,13 @@ hokulea-registry = { path = "crates/registry", version = "0.1.0", default-featur
 eigenda-cert = { path = "crates/eigenda-cert" }
 
 # Alloy (Network)
-alloy-consensus = { version = "=1.6.3", default-features = false }
-alloy-genesis = { version = "=1.6.3", default-features = false }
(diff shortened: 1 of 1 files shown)


fakedev9999:
Closing — relaxing the alloy exact pins causes hokulea to resolve alloy 1.7.x which breaks compilation (hokulea code depends on alloy 1.6.x APIs). The exact pins are intentional. Will use the fork rev (af7f05e, based on v1.1.6 before the alloy pin) in op-succinct until a release tag is cut.

PR #316, opened 2026-05-13
chore(deps): bump kona to kona-client/v1.2.14

fakedev9999:
Bumps kona git refs from `kona-client/v1.2.13` to `kona-client/v1.2.14`.

`kona-client/v1.2.14` is a fix-only backport of the DestroyedChanged account fix in `kona-executor` (GHSA-6gcx-mhp9-fxjf). Diff between v1.2.13 and v1.2.14 is a single file (`rust/kona/crates/proof/executor/src/db/mod.rs`: 4-line fix + 5 regression tests), no Cargo.toml/Cargo.lock changes in kona itself.

Workspace-level Cargo.lock churn (source URL swaps for alloy-op-evm, op-alloy, etc.) caused the canoe-sp1-cc-client ELF to rebuild non-trivially (~2.5 MB byte diff), so `V_KEY` in `canoe/sp1-cc/verifier/src/lib.rs` is updated accordingly.

Built with `cargo prove build --docker --tag v6.0.1`.


bxue-l2:
Thanks for the PR. One question, why built with v6.0.1 as opposed to v6.1.0, which is currently used as the dep. https://github.com/Layr-Labs/hokulea/blob/b521561c763f48a896774e0648e7b5ef68944906/Cargo.toml#L140

fakedev9999:
Followed `.github/workflows/elf.yml` which currently uses `--tag v6.0.1`. I noticed the workspace pins `sp1-hypercube = "=6.1.0"` and `justfile` defaults to v6.1.0, so I rebuilt with `--tag v6.1.0` to verify — ELF is byte-identical and V_KEY is the same. No PR change needed.

Probably worth aligning `elf.yml` to v6.1.0 in a separate PR to remove the discrepancy.

fakedev9999:
Pushed `aff8b7d` to align `elf.yml` to v6.1.0 in this PR. v6.2.0 bump deferred to a separate cascade (rsp + sp1-cc + hokulea).

bxue-l2:
good catch, I will update the elf workflow version.

bxue-l2 (approved):

fakedev9999:
Re-signed all 3 commits — `fa009b1`, `e26cb4e`, `b76a7af` all show Verified now. Force-pushed. Ready to merge.

PR #318, opened 2026-05-18
chorus: add eigenda witness from preimage unchecked and debug lines

bxue-l2:
Summary

 1. EigenDAWitness::from_preimage_unchecked (ffb8271)
  - New constructor on the witness type that builds it directly from a preimage without re-validating. Touches crates/proof/src/eigenda_witness.rs and the witgen call sites (canoe_witness_provider.rs,
  witgen/src/lib.rs).

  2. Narrow from_eigenda_preimage_to_canoe_inputs signature (7fc4f04, 8d71b10)
  - Function now takes validities directly instead of deriving them, simplifying the caller contract.

  3. Drop rsp-primitives from canoe-sp1-cc (2e4b193, caf7fef, a5745a1)
  - Inlined genesis_from_json into canoe/sp1-cc/host/src/lib.rs so the host and verifier no longer depend on rsp-primitives. Removed from both crate Cargo.tomls and the workspace root.

  4. Debug instrumentation (c3b2cb2, 1d34001, 0b15739)
  - Added blob-size logging at proof generation in crates/compute-proof/src/kzg_proof.rs

### DIFF
--- a/crates/compute-proof/src/kzg_proof.rs
+++ b/crates/compute-proof/src/kzg_proof.rs
@@ -38,12 +38,11 @@ pub fn compute_kzg_proof_with_srs<'s>(
 
     let commit_start = Instant::now();
     let commitment = kzg.commit_eval_form(&input_poly, srs)?;
-
     debug!(
         target: "compute_kzg_proof_with_srs",
-        "blob size {:?} bytes. Commit in eval form takes {:?}",
-        encoded_payload.len(),
-        commit_start.elapsed(),
+        blob_size_bytes = encoded_payload.len(),
+        elapsed = ?commit_start.elapsed(),
+        "Commit in eval form"
     );
 
     let proof_start = Instant::now();
@@ -51,9 +50,9 @@ pub fn compute_kzg_proof_with_srs<'s>(
 
     debug!(
         target: "compute_kzg_proof_with_srs",
-        "blob size {:?} bytes. Proof generations takes {:?}",
-        encoded_payload.len(),
-        proof_start.elapsed(),
+        blob_size_bytes = encoded_payload.len(),
+        elapsed = ?proof_start.elapsed(),
+        "Proof generation"
     );
 
     let proof_x_bigint: BigUint = proof.x.into();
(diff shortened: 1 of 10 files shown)


crates/compute-proof/src/kzg_proof.rs
```diff
@@ -26,16 +28,34 @@ pub fn compute_kzg_proof_with_srs<'s>(
...
 
+    let commit_start = Instant::now();
     let commitment = kzg.commit_eval_form(&input_poly, srs)?;
 
+    debug!(
+        target: "compute_kzg_proof_with_srs",
+        "blob size {:?} bytes. Commit in eval form takes {:?}",
+        encoded_payload.len(),
+        commit_start.elapsed(),
+    );
```
kien-rise, line 46: I recommend using [fields](https://docs.rs/tracing/latest/tracing/index.html#recording-fields):

```rust
debug!(
    target: "compute_kzg_proof_with_srs",
    blob_size_bytes = encoded_payload.len(),
    elapsed = ?commit_start.elapsed(),
    "Commit in eval form"
);
```

crates/compute-proof/src/kzg_proof.rs
```diff
@@ -26,16 +28,34 @@ pub fn compute_kzg_proof_with_srs<'s>(
...
+
+    let proof_start = Instant::now();
     let proof = kzg.compute_blob_proof(&blob, &commitment, srs)?;
+
+    debug!(
+        target: "compute_kzg_proof_with_srs",
+        "blob size {:?} bytes. Proof generations takes {:?}",
+        encoded_payload.len(),
+        proof_start.elapsed(),
+    );
```
kien-rise, line 56:
```rust
debug!(
    target: "compute_kzg_proof_with_srs",
    blob_size_bytes = encoded_payload.len(),
    elapsed = ?proof_start.elapsed(),
    "Proof generation"
);
```

kien-rise (approved):
I made some comments but they are not critical.

crates/compute-proof/src/kzg_proof.rs
bxue-l2: 1d34001b339b6c0d43ed396d628e820d841bdbe5

crates/compute-proof/src/kzg_proof.rs
bxue-l2: 1d34001b339b6c0d43ed396d628e820d841bdbe5

PR #320, opened 2026-05-28
feat: add sp1-bn-verifier for optimized kzg batch verification

bxue-l2:
## Summary

  Adds an opt-in substrate-bn (sp1-patches) batch KZG verifier as an alternative to the default arkworks backend. Inside an SP1 zkVM the patched bn primitives win big — one user reports `from_witness()` sample share fell from **65% → 22%**
  (`343,329 → 53,345`, **−74%**). Also includes memory-allocation cleanups in `from_witness` that benefit the default ark path too.

  ## What changed

  - **New crate `hokulea-sp1-bn-verifier`** — `no_std` port of `verify_blob_kzg_proof_batch` on `sp1-patches/bn`. Byte-identical Fiat-Shamir transcript to the arkworks reference, asserted by the parity suite in
  `crates/sp1-bn-verifier/tests/parity.rs`.
  - **Feature-gated swap at the call site** in `crates/proof/src/preloaded_eigenda_provider.rs::batch_verify`. `--features ark` (default) → arkworks; `--features sp1-bn` → substrate-bn. Both on: `sp1-bn` wins. Neither on: `compile_error!`.
  - **Optional arkworks deps** — `rust-kzg-bn254-{primitives,verifier}`, `ark-bn254`, `ark-ff` are gated behind `ark`, so `--features sp1-bn` doesn't pull them in directly.
  - **Passthrough features** on every consumer (`bin/host`, `bin/client`, `crates/zkvm-verification`, `crates/compute-proof`, `example/preloader`) so the backend is picked at the consumer level.
  - **Empty-batch short-circuit** — `substrate-bn`'s `AffineG1::msm` panics on a zero-length slice; the new impl returns `Ok(true)` early to match the reference verifier (after the length-equality guard, so mismatched empty inputs still error).
  - **`from_witness` allocation cleanup** — `validity_entries` via `into_iter().collect()` (no clones); `encoded_payload_entries` populated only on success; `PolynomialEvalForm::new` computed once per blob instead of 3× (the upstream library has
  the same redundancy with a TODO); slice → iterator API removes intermediate `Vec` materializations.

  ## Usage

  ```sh
  # Default (arkworks)
  just run-preloader .devnet.env

  # Substrate-bn (sp1-patches)
  just run-preloader .devnet.env sp1-bn

  For host/client/library builds use --no-default-features --features sp1-bn to skip compiling unused arkworks deps. Confirm the feature wired through with:

  cargo tree -p hokulea-example-preloader --features sp1-bn -e features | grep hokulea-sp1-bn-verifier
```

Thanks @kien-rise for initiation, testing, contribution and feedback on this feature

### CODE REVIEW

File: crates/sp1-bn-verifier/Cargo.toml [hunk 1]
```diff
@@ -0,0 +1,26 @@
...
+[dev-dependencies]
+ark-bn254.workspace = true
+ark-ff.workspace = true
+rust-kzg-bn254-primitives.workspace = true
+rust-kzg-bn254-prover.workspace = true
+rust-kzg-bn254-verifier.workspace = true
+hokulea-compute-proof.workspace = true
+hokulea-eigenda.workspace = true
+num.workspace = true
+substrate-bn = { git = "https://github.com/sp1-patches/bn", tag = "patch-0.6.0-sp1-6.0.0-substrate-bn" }
```
kien-rise: nit: deps already included in `[dependencies]` can be excluded in `[dev-dependencies]`

crates/witgen/Cargo.toml
```diff
@@ -11,7 +11,7 @@ alloy-rlp.workspace = true
 
 kona-proof.workspace = true
 kona-preimage.workspace = true
-hokulea-proof.workspace = true
+hokulea-proof = { workspace = true, features = ["ark"] }
```
kien-rise: I missed this one. Let's apply the feature forwarding technique here for consistency:

```
[features]
default = ["ark"]
ark = ["hokulea-proof/ark"]
sp1-bn = ["hokulea-proof/sp1-bn"]
```

bxue-l2: good catch

bxue-l2: https://github.com/Layr-Labs/hokulea/pull/320/changes/8495d7268833c1555f2e011f7a951ac367bb8c8c

crates/sp1-bn-verifier/Cargo.toml
bxue-l2: https://github.com/Layr-Labs/hokulea/pull/320/changes/8495d7268833c1555f2e011f7a951ac367bb8c8c

PR #324, opened 2026-06-09
chore(deps): bump op-stack to op-reth/v2.3.1 (reth v2.2.0)

seolaoh:
Bumps the op-stack dependencies from `kona-client/v1.5.2` to `op-reth/v2.3.1` (paradigmxyz/reth v2.2.0).

Rebased on top of the merged #325 (kona to 1.5.2), so it is now a Cargo.toml-only change:

- The 12 kona workspace deps and the 9 `[patch.crates-io]` op-alloy / alloy-op / op-revm entries: retagged from `kona-client/v1.5.2` to `op-reth/v2.3.1`.
- The five exact alloy network pins (`alloy-consensus`, `alloy-genesis`, `alloy-provider`, `alloy-rpc-client`, `alloy-rpc-types`): `=2.0.4` to `=2.0.5`.

The kona crate versions are identical at both tags, so no code changes are needed; only alloy core/primitives differ (`alloy-core` 1.5.7 to 1.6.0). `Cargo.lock` is updated to match.

Why: celo-kona feeds both celo-reth (the L2 node) and the proving stack, and cargo unifies git deps by source (URL + tag), so hokulea and celo-kona have to point at the same optimism tag. celo-reth is moving to `op-reth/v2.3.1`, so an `op-reth/v2.3.1` build of hokulea available upstream lets us share a single dependency graph.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -135,9 +135,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus"
-version = "2.0.4"
+version = "2.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e3d64da86c616b5092ea64eea648f311bbd58630a0b384c42d699175d6f9122b"
+checksum = "83447eeb17816e172f1dfc0db1f9dc0b7c5d069bd1f7cecbecceb382bf931015"
 dependencies = [
  "alloy-eips 2.0.5",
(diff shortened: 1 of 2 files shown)


bxue-l2:
created a branch at https://github.com/Layr-Labs/hokulea/tree/hokulea-client/op-reth/v2.3.1

PR #327, opened 2026-06-19
chore(deps): bump kona to v1.6.0

fakedev9999:
Bumps kona from `kona-client/v1.5.2` to `v1.6.0` (the latest stable release). Same crate versions, so it's a tag-only change with no code changes. The canoe sp1-cc ELF and vkey are unaffected — kona isn't a dependency of that guest.


fakedev9999:
Superseded by #328

